PR #177, opened 2025-11-14
Replace CRLF with LF when loading file

### LINKED ISSUE #176: Updates are broken on files using CRLF line terminators
## Description

Updates are broken on files using CRLF line terminators (`\r\n` instead of `\n`). This generally only appears in files created on Windows systems.

https://github.com/user-attachments/assets/46443a8b-22c6-4f43-ab2d-4d7da71864a9

This bug also happens on Jupyter Collaboration v2, v3, v4. The root cause of this bug likely lives in `y-codemirror`: https://github.com/yjs/y-codemirror

## Reproduce

1. Download this file & upload it to JupyterLab: 

[testing_file.py](https://github.com/user-attachments/files/23551280/testing_file.py)

2. Run the script and verify it works as expected (printing `500` and `650`)

3. Change the argument on the last line

4. Observe that the script breaks

dlqqq:
## Description

- Fixes #176.

## Demo

https://github.com/user-attachments/assets/6c9fd36a-aec3-4f67-9775-9445ac52be2f

### DIFF
--- a/jupyter_server_documents/rooms/yroom_file_api.py
+++ b/jupyter_server_documents/rooms/yroom_file_api.py
@@ -321,8 +321,8 @@ class YRoomFileAPI(LoggingConfigurable):
 
         # Replace CRLF line terminators with LF line terminators
         # Fixes #176, see issue description for more context.
-        content: str = file_data['content']
-        if '\r\n' in content:
+        content = file_data.get('content')
+        if isinstance(content, str) and '\r\n' in content:
             self.log.warning(f"Detected CRLF line terminators in '{path}'.")
             content = content.replace('\r\n', '\n')
             self.log.info("Replaced CRLF line terminators with LF line terminators.")


jupyter_server_documents/rooms/yroom_file_api.py
```diff
@@ -319,9 +319,17 @@ async def _load_content(self, jupyter_ydoc: YBaseDoc) -> None:
             self.log.info(f"Processing outputs for loaded notebook: '{self.room_id}'.")
             file_data = self.outputs_manager.process_loaded_notebook(file_id=self.file_id, file_data=file_data)
 
+        # Replace CRLF line terminators with LF line terminators
+        # Fixes #176, see issue description for more context.
+        content: str = file_data['content']
```
3coins: Is this guaranteed to be always str?

dlqqq: Good catch. I think this should always be a string, but there may be some weird scenario where this is an instance of `bytes`.

I can use the `isinstance()` function to prevent any weird edge case.

dlqqq: Lemme work on that now

3coins (approved):
Looks good!

PR #180, opened 2025-11-23
Remove save toolbar buttons

### LINKED ISSUE #179: Empty Save Icon
We disable the "Save" icon—which we should—but it still appears as an empty (clickable!) icon. It should be removed altogether.

Image

nakul-py:
Removing Ghost "Save" toolbar button.

Screenshot From 2025-11-23 18-05-03

- Closes #179 

cc @Zsailer @dlqqq


dlqqq:
Demo screenshot of new CSS implementation:

Screenshot 2026-01-13 at 4 55 32 PM

PR #182, opened 2025-12-03
WIP: Updates copy cell command to exclude outputs

3coins:


dlqqq:
I don't think this is needed anymore since the outputs service is disabled by default now. It needs a rebase anyways so the best thing to do would be to open a new PR to revive this.

PR #184, opened 2025-12-05
Fix metadata links

ktaletsk:
## Summary
- fix the package homepage and bugs links so they point to the actual hyphenated repository name


dlqqq (approved):
@ktaletsk Thank you!

PR #200, opened 2026-03-20
Lower WS ping interval to 25 seconds

### LINKED ISSUE #199: YRoomWebsocket has no ping/pong keepalive, causing 1006 disconnections behind proxies
## Summary

`YRoomWebsocket` has no WebSocket ping/pong keepalive. Idle connections are killed by reverse proxies and load balancers, resulting in close code 1006.

## Details

`YRoomWebsocket` inherits directly from Tornado's `WebSocketHandler`, which defaults `ping_interval` to `None` (disabled) ([tornado/websocket.py#L295](https://github.com/tornadoweb/tornado/blob/v6.5.4/tornado/websocket.py#L295)). `None` resolves to `0` at the connection layer ([tornado/websocket.py#L1316-L1322](https://github.com/tornadoweb/tornado/blob/v6.5.4/tornado/websocket.py#L1316-L1322)), so no pings are ever sent.

Jupyter Server's `ServerApp.websocket_ping_interval` and `websocket_ping_timeout` are `Integer` traits with no default set, which means they default to `0` ([serverapp.py#L1652-L1676](https://github.com/jupyter-server/jupyter_server/blob/v2.17.0/jupyter_server/serverapp.py#L1652-L1676)). These are passed straight through to Tornado's settings dict with no conversion ([serverapp.py#L456-L457](https://github.com/jupyter-server/jupyter_server/blob/v2.17.0/jupyter_server/serverapp.py#L456-L457)). So unless a user explicitly configures them, pings are disabled for any handler that relies on Tornado's native keepalive.

Common reverse proxies and load balancers will kill idle WebSocket connections:

- **Nginx**: `proxy_read_timeout` defaults to 60 seconds ([nginx.org docs](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_read_timeout))
- **AWS ALB**: idle timeout defaults to 60 seconds ([AWS docs](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/edit-load-balancer-attributes.html))
- **GCP load balancers**: backend service timeout defaults to 30 seconds ([GCP docs](https://cloud.google.com/load-balancing/docs/backend-service#timeout))

The kernel WebSocket (`KernelWebsocketHandler`) logs a warning about `websocket_ping_timeout (90000) cannot be longer than websocket_ping_interval (30000)` — this is a cosmetic issue caused by `jupyter_server`'s `WebSocketMixin` properties (in milliseconds) shadowing Tornado's properties (in seconds) via MRO. The kernel WS still has a working 30-second ping loop from `WebSocketMixin`'s custom `PeriodicCallback`, so it's not actually broken. This is an upstream `jupyter_server` issue.

## Fix

Set `ping_interval` and `ping_timeout` as class attributes on `YRoomWebsocket` to enable Tornado's native keepalive:

```python
class YRoomWebsocket(WebSocketHandler):
    ping_interval = 25  # seconds
    ping_timeout = 10   # seconds
```

A 25-second interval keeps connections alive through proxies with 30–60 second idle timeouts.

dlqqq:
Sets `websocket_ping_interval=25` and `websocket_ping_timeout=10` (in seconds) in `_link_jupyter_server_extension`.

Previously these values were configured as 30000/90000, which Tornado interprets as ~8.3/~25 hours since it expects seconds. This effectively disabled the ping/pong keepalive, causing WebSocket 1006 disconnections behind reverse proxies and load balancers (Nginx 60s default timeout, AWS ALB 60s default idle timeout).

A 25-second ping interval keeps connections alive through proxies with 30–60 second idle timeouts, and the 10-second pong timeout allows prompt detection of dead connections.

Fixes #199

### DIFF
--- a/jupyter_server_documents/app.py
+++ b/jupyter_server_documents/app.py
@@ -82,6 +82,8 @@ class ServerDocsApp(ExtensionApp):
     def _link_jupyter_server_extension(self, server_app):
         """Setup custom config needed by this extension."""
         c = Config()
+        c.ServerApp.websocket_ping_interval = 25
+        c.ServerApp.websocket_ping_timeout = 10
         c.ServerApp.kernel_websocket_connection_class = "jupyter_server_documents.kernels.websocket_connection.NextGenKernelWebsocketConnection"
         c.ServerApp.kernel_manager_class = "jupyter_server_documents.kernels.multi_kernel_manager.NextGenMappingKernelManager"
         c.MultiKernelManager.kernel_manager_class = "jupyter_server_documents.kernels.kernel_manager.NextGenKernelManager"


dlqqq:
See update in #199.

PR #202, opened 2026-03-21
 Fix cell duplication on out-of-band file changes by reloading file content in-place

sathishlxg:
### What

When a file was modified out-of-band (outside of Jupyter), the server detected the last_modified timestamp change and responded by restarting the YRoom (handle_outofband_change → restart(close_code=4000)). This disconnected all clients and reloaded the file into a fresh YDoc with no history.

When clients reconnected and performed the SyncStep1/SyncStep2 handshake, the CRDT merge between the client's local YDoc (which still had the original history) and the server's newly initialized YDoc (loaded from the updated file with no shared history) produced duplicate cells — the client's existing cells and the server's reloaded cells were both retained by the merge.

  #### Fix

Instead of restarting the room, out-of-band changes now trigger **_reload_content_inplace()**, which applies the updated file content directly to the existing live YDoc. Because the YDoc's history is preserved, **pycrdt** correctly computes the diff and broadcasts it to connected clients as a normal SyncUpdate. Clients receive the change without disconnecting or diverging state, eliminating the duplication.

To prevent the reloaded content from being immediately saved back to disk (which would update last_modified and re-trigger the change detection), a **_reloading_content** flag suppresses schedule_save() in **_on_jupyter_ydoc_update** for the duration of the YDoc write.

##### Changes

- **yroom_file_api.py:**
    - Add _reload_content_inplace(jupyter_ydoc) — applies updated file content to the live YDoc in-place, keeping clients connected
    - _check_file() now takes jupyter_ydoc and calls _reload_content_inplace() instead of handle_outofband_change() on timestamp mismatch
    - Add _reloading_content flag to suppress saves triggered by the reload write
 - **yroom.py:**
    - _on_jupyter_ydoc_update skips schedule_save() when file_api._reloading_content is True

https://github.com/user-attachments/assets/e0155464-1e0e-4c02-90d7-540ca5c7b756

### DIFF
--- a/jupyter_server_documents/rooms/yroom.py
+++ b/jupyter_server_documents/rooms/yroom.py
@@ -710,7 +710,7 @@ class YRoom(LoggingConfigurable):
         # Do nothing if an in-place reload is in progress. The update was
         # triggered by reading a new version of the file from disk, so saving
         # it back would cause an infinite reload loop via last_modified checks.
-        if self.file_api._reloading_content:
+        if self.file_api.reloading_content:
             return
 
         # Do nothing if the event updates the 'state' dictionary with no effect
@@ -819,17 +819,6 @@ class YRoom(LoggingConfigurable):
         out-of-band changes.
         """
         self.restart(close_code=4000, immediately=True)
-
-        
-    def handle_outofband_change(self) -> None:
-        """
-        Handles an out-of-band change by restarting the YRoom immediately,
-        closing all Websockets with close code 4000.
-
-        See `restart()` for more info.
-        """
-        self.restart(close_code=4000, immediately=True)
-    
 
     def handle_outofband_move(self) -> None:
         """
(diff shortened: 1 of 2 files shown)


sathishlxg:
@3coins I confirmed the issue also exists in versions earlier than 4.5.3. Duplicate cells are being added to the shared model and rendered in the DOM, but they remain hidden in the UI due to a bug in deferred cell rendering. Because these cells are duplicated, executing the Run All command also runs the hidden duplicates. While the current fix—removing and re-adding the cells—works for simple cases, it may introduce performance issues for large notebooks with frequent out-of-band iterative changes.

image

Zsailer:
Yeah, let's wait for @dlqqq to review before merging, since he's thought the most about this out-of-band challenge. 

I believe there were some edge cases that made reloading the file in-place inviable.

PR #207, opened 2026-03-27
Clears outputs on each execute

### LINKED ISSUE #194: Notebook outputs occasionally duplicated, possibly after document reset
While testing PR #193, I noticed that notebook outputs are occasionally duplicated after a room is freed and then re-opened:

Image

While this could be a bug introduced by #193, other testers have seen the same issue in Jupyter AI, when an agent makes out-of-band edits to a notebook. I believe this is tied to document resets / room restarts.

Couple of other observations:

1. The duplicate outputs are being saved to the notebook - verified via `cat *.ipynb`
2. This occurs with default config, i.e. with `OutputsManager` (have not tested with `OptimizedOutputsManager`)
3. The output gets duplicated once after the room is restarted AND I re-run the cell. So if there are 2 copies of the same output, if I restart the notebook and run that cell, I get 3 copies when I re-open the notebook again.

### LINKED ISSUE #206: Outputs accumulate when a cell is executed with a new kernel session
https://github.com/user-attachments/assets/0c166756-38fe-4382-b741-cf0babed238a

### Steps to reproduce
1. Create a new notebook, open in JupyterLab and execute a cell, close the notebook
2. Either shut down the kernel or restart JupyterLab
3. Open the notebook and execute the same cell again
4. While the UI will show 1 output, if you look at the saved notebook file, it will show 2 outputs
5. If you refresh the browser window, you can see the 2 outputs

### Root cause
Cell outputs are only being cleared when the cell is re-executed, this is dependent on the `message_cache` in `kernel_client.py`, which is an in-memory cache and gets reset when new kernel session is started.
https://github.com/jupyter-ai-contrib/jupyter-server-documents/blob/4272ef69ab1f5b9b7942c51bfecb52577d114d64/jupyter_server_documents/kernels/kernel_client.py#L127-L130 

Consequently, the first cell execution after a new kernel session doesn't clear the outputs, since the cache is empty.

3coins:
Fixes #206 
Fixes #194

### DIFF
--- a/jupyter_server_documents/kernels/kernel_client.py
+++ b/jupyter_server_documents/kernels/kernel_client.py
@@ -123,15 +123,12 @@ class DocumentAwareKernelClient(AsyncKernelClient):
         metadata = self.session.unpack(msg[2])
         cell_id = metadata.get("cellId")
 
-        # Clear cell outputs if cell is re-executed
-        if cell_id:
-            existing = self.message_cache.get(cell_id=cell_id)
-            if existing and existing['msg_id'] != msg_id:
-                asyncio.create_task(self.output_processor.clear_cell_outputs(cell_id))
-
         # IMPORTANT: Set cell to 'busy' immediately when execute_request is received
         # This ensures queued cells show '*' prompt even before kernel starts processing them
+        # Also clear outputs immediately for this cell execution
         if msg_type == "execute_request" and channel_name == "shell" and cell_id:
+            # Always clear outputs when executing a cell
+            asyncio.create_task(self.output_processor.clear_cell_outputs(cell_id))
             for yroom in self._yrooms:
                 yroom.set_cell_awareness_state(cell_id, "busy")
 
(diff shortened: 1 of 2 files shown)


dlqqq (approved):
Looks great, thanks for working on this. Left a few non-blocking comments for you to consider below.

jupyter_server_documents/kernels/kernel_client.py
```diff
@@ -123,15 +123,12 @@ def handle_incoming_message(self, channel_name: str, msg: list[bytes]):
...
-            existing = self.message_cache.get(cell_id=cell_id)
-            if existing and existing['msg_id'] != msg_id:
-                asyncio.create_task(self.output_processor.clear_cell_outputs(cell_id))
-
         # IMPORTANT: Set cell to 'busy' immediately when execute_request is received
         # This ensures queued cells show '*' prompt even before kernel starts processing them
+        # Also clear outputs immediately for this cell execution
         if msg_type == "execute_request" and channel_name == "shell" and cell_id:
+            # Always clear outputs when executing a cell
+            asyncio.create_task(self.output_processor.clear_cell_outputs(cell_id))
```
agoose77, line 131: @dlqqq just browsing over this cursorily — is there any other synchronisation to avoid a race condition here (given that we don't await the task)?

PR #215, opened 2026-04-09
Fix content duplication on reconnection

### LINKED ISSUE #210: Data duplication when re-connecting
Follow-up to #209.

https://github.com/user-attachments/assets/bea4538b-a245-40c9-b28d-812f6e3c0db1

## Hypothesis

I think this only happens re-connecting after restarting the server because the server has a new YDoc. So the client's copy gets merged with the server's copy, resulting in data duplication. We probably need to port over Jupyter Collaboration's concept of a session ID (a random UUID set once per server session). 

## Proposed fix

If client's session ID doesn't match the server's session ID, then the server should close the WS with a special code (e.g. 4004) and then the client should retry after deleting its shared model. 

## Notes

However Brian reports that he was seeing data duplication even without restarting the server. We need more investigation to confirm.

dlqqq:
## Demo

https://github.com/user-attachments/assets/1bc23453-415a-4839-b122-b736f83974fb

## Description

Fixes #210 

Fixes the content duplication bug that occurs in the two currently known scenarios:

- Reconnecting after restarting the server
- Reconnecting after a room was freed from memory (can be caused by closing your laptop for an hour while connected to a remote server)

In both cases, the duplication occurs because the client is connecting with an existing YDoc that has the same content but a divergent history. If the server's YDoc has a different history (e.g. it contains only a single item when it loads from disk), the histories are merged even though the content is identical, resulting in content duplication.

### Divergent history detection

To fix this, this PR implements **divergent history detection**. The `YRoom` now checks the state vector provided by each client when connecting to the `YRoom`. There are 3 scenarios:

1. State vector is totally empty: client just opened the file in their browser for first time
2. State vector is non-empty but has the same set of Yjs clients => client briefly disconnected but has the same history
3. **State vector has a different set of Yjs clients** => client has **divergent history** from a previous room session

This allows us to precisely detect when content duplication would occur.

### How content duplication is prevented

When case 3) occurs, the `YRoom` does the following to prevent content duplication:

1. Clear the YDoc to an empty state, while storing the original state
2. Queue all document updates and block saves
3. Perform the handshake as normal
4. Restore the YDoc to its original state
5. Broadcast all queued document updates and resume saving

- Step 1) prevents the reconnecting client from seeing content duplication.
- Step 2) prevents other clients from seeing an empty document while the reconnecting client is completing the handshake
- Step 3) syncs the reconnecting client
- Steps 4) + 5) gets all clients synced to the original document state as quickly as possible

This logic runs only if a connecting client has divergent history. It does not run in the "normal" circumstances (fresh connection / reconnecting after brief disconnect).

###  Consequences

- Edits made during the brief window where a client with divergent history is connecting may appear in the wrong locations.
- Other clients may see a brief flash of an empty document. This timing window has been minimized thanks to the new `YRoomUpdateBuffer`, but it may be visible very briefly on the browser.

### Additional code changes

Each sync handshake now blocks the processing of new messages until it is fully complete.

## Notes

While this solves the content duplication issue, I do not think this is the right long-term solution. We're essentially freezing edits every time a divergent client connects for every other client. This also requires clients to sync one-at-a-time since we must wait for the divergent client's SS2 reply - this can cause issues if several people try to connect to a room at the same time (e.g. in a teaching environment).

The ideal solution would be to have each **client** do the divergent history detection based on the SS1 message sent from the server after the first handshake, then reset their YDoc to a completely empty state if the client discovers that the server has diverged. Previously this is how we handled out-of-band changes. However, we discovered that this was causing UI state to get corrupted, which motivated #202.

Long-term, the right solution is to work to improve the reliability of resetting the YDoc to an empty state in the JupyterLab UI. Essentially #93 needs to work in JupyterLab without corrupting the UI - that would allow us to handle divergent clients more reliably.

### DIFF
--- a/jupyter_server_documents/rooms/yroom.py
+++ b/jupyter_server_documents/rooms/yroom.py
@@ -238,7 +238,9 @@ class YRoom(LoggingConfigurable):
         self._on_stop_callbacks: list[Callable[[], Any]] = []
         self._stopped = False
         self._updated = False
-        self._pending_source_restore: dict[str, str] = {}
+        self._pending_ss2_future: asyncio.Future[bytes] | None = None
+        self._pending_ss2_client_id: str | None = None
+        self._suppress_broadcasts = False
         self._save_task = None
         self._last_activity = time.monotonic()
         self.show_gc_debug = self.parent.show_gc_debug
@@ -519,7 +521,21 @@ class YRoom(LoggingConfigurable):
         """
         Adds new message to the message queue. Items placed in the message queue
         are handled one-at-a-time.
+
+        If handle_sync_step1 is awaiting an SS2 reply from a client, the reply
+        bypasses the queue and resolves the pending future directly.
         """
+        if (
+            self._pending_ss2_future is not None
+            and not self._pending_ss2_future.done()
+            and client_id == self._pending_ss2_client_id
+            and len(message) >= 2
+            and message[0] == YMessageType.SYNC
+            and message[1] == YSyncMessageSubtype.SYNC_STEP2
+        ):
+            self._pending_ss2_future.set_result(message)
+            return
+
         self._message_queue.put_nowait((client_id, message))
     
 
@@ -549,7 +565,7 @@ class YRoom(LoggingConfigurable):
 
             # Otherwise, process & handle the new message
             client_id, message = queue_item
-            self.handle_message(client_id, message)
+            await self.handle_message(client_id, message)
             
             # Finally, inform the asyncio Queue that the task was complete
             # This is required for `self._message_queue.join()` to unblock once
@@ -561,16 +577,21 @@ class YRoom(LoggingConfigurable):
             f"for YRoom '{self.room_id}'."
         )
     
-    def handle_message(self, client_id: str, message: bytes) -> None:
+    async def handle_message(self, client_id: str, message: bytes) -> None:
         """
         Handles all messages from every client received in the message queue by
         calling the appropriate handler based on the message type. This method
         routes the message to one of the following methods:
 
-        - `handle_sync_step1()`
-        - `handle_sync_step2()`
-        - `handle_sync_update()`
-        - `handle_awareness_update()`
+        - `await handle_sync()`: performs a sync handshake initiated by a
+        client's SS1 message and returns once the handshake is complete. This
+        blocks handling of other messages until the sync handshake is complete.
+
+        - `handle_sync_update()`: immediately applies a document update from a
+        synced client.
+
+        - `handle_awareness_update()`: immediately applies an awareness update
+        from a synced client.
         """
 
         # Determine message type & subtype from header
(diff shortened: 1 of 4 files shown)


3coins:
Not sure if it's related to the changes in the PR, but I still see the dirty flag not reset on re-connect. This is happening on remote servers, have not seen with server running on my laptop. 
 
Screenshot 2026-04-10 at 10 57 55 AM

Here are the logs:

```
[I 2026-04-10 17:55:44.803 ServerDocsApp] Initializing room 'json:notebook:8e6707b4-20fb-4921-9882-5c04e8452ac2'.
[I 2026-04-10 17:55:44.804 ServerDocsApp] Room 'json:notebook:8e6707b4-20fb-4921-9882-5c04e8452ac2' initialized.
[I 2026-04-10 17:55:44.805 ServerDocsApp] Loading content for room ID 'json:notebook:8e6707b4-20fb-4921-9882-5c04e8452ac2', found at path: 'test-1.ipynb'.
[I 2026-04-10 17:55:44.805 ServerDocsApp] Processing outputs for loaded notebook: 'json:notebook:8e6707b4-20fb-4921-9882-5c04e8452ac2'.
[I 2026-04-10 17:55:44.856 ServerDocsApp] Loaded content for room ID 'json:notebook:8e6707b4-20fb-4921-9882-5c04e8452ac2'.
[W 2026-04-10 17:55:44.858 ServerDocsApp] Client 'fa2c8d4b-463c-453f-a428-cd3be6e96580' has divergent history in room 'json:notebook:8e6707b4-20fb-4921-9882-5c04e8452ac2'. Clearing YDoc source before handshake.
[I 2026-04-10 17:55:44.860 ServerDocsApp] Initiating handshake with client 'fa2c8d4b-463c-453f-a428-cd3be6e96580' in room 'json:notebook:8e6707b4-20fb-4921-9882-5c04e8452ac2'.
[I 2026-04-10 17:55:44.943 ServerDocsApp] Completed handshake with client 'fa2c8d4b-463c-453f-a428-cd3be6e96580' in room 'json:notebook:8e6707b4-20fb-4921-9882-5c04e8452ac2'.
[I 2026-04-10 17:55:44.943 ServerDocsApp] Restored YDoc source.
[I 2026-04-10 17:55:46.864 ServerDocsApp] Initiating handshake with client '798aeb8e-2a17-4a79-b6df-9e63e69bbc6c' in room 'json:notebook:8e6707b4-20fb-4921-9882-5c04e8452ac2'.
[I 2026-04-10 17:55:46.934 ServerDocsApp] Completed handshake with client '798aeb8e-2a17-4a79-b6df-9e63e69bbc6c' in room 'json:notebook:8e6707b4-20fb-4921-9882-5c04e8452ac2'.
```

3coins:
I can consistently reproduce the dirty flag issue now, if 2 clients are connected and the user(s) try to reconnect one after another in quick succession. 

This can be reproduced by following these steps:

1. Run the server on a remote machine
2. Open JupyterLab in 2 browsers side-by-side
3. Shutdown the server, this should bring the reconnect dialog.
4. Start the server again, and reconnect in both browsers.
5. The client that you reconnected first will show the dirty flag.

dlqqq:
Thanks for reporting this issue. I was able to reproduce the issue on this branch but also on the `main`  branch. Seems like this is an existing issue.

Issue #216, opened 2026-04-10
IndexError in OutputProcessor.output_task() when clear_output races with update_display_date


j1wonpark:
## Description 

`OutputProcessor.output_task()` raises `IndexError: Array index out of range` when a `clear_output` message and and  `update_display_data`message are processed concurrently.

## Steps to reproduce

1. Use any Jupyter extension or code that class `display(HTML(...), display_id=True)`followed by `handle.update(HTML(...))`repeatedly
2. Re-execute the cell or trigger `clear_output`while updates are in-flight

## Traceback

```
Task exception was never retrieved
future: <Task finished name='Task-1873' coro=<OutputProcessor.output_task() done,
  defined at jupyter_server_documents/outputs/output_processor.py:141>
  exception=IndexError('Array index out of range')>
Traceback (most recent call last):
  File "jupyter_server_documents/outputs/output_processor.py", line 181, in output_task
    target_cell["outputs"][output_index] = output
  File "pycrdt/_array.py", line 218, in __setitem__
    key = self._check_index(key)
  File "pycrdt/_array.py", line 240, in _check_index
    raise IndexError("Array index out of range")
IndexError: Array index out of range
```

PR #217, opened 2026-04-10
fix: bounds-check output index before array assignment (#216)

j1wonpark:
## Summary
- Fix `IndexError: Array index out of range` in `OutputProcessor.output_task()` when `clear_output` races with `update_display_data` (#216)
- Add bounds check on `output_index` before writing to `target_cell["outputs"]`, falling back to `append` when the index is out of range

## Test plan
- [x] Added `test_output_task_update_display_data` — verifies display_id output replacement works
- [x] Added `test_output_task_update_display_after_clear_no_index_error` — regression test reproducing #216 race condition
- [x] Both tests fail with `IndexError` before the fix, pass after
- [x] Full test suite passes (159 tests)

### DIFF
--- a/jupyter_server_documents/outputs/output_processor.py
+++ b/jupyter_server_documents/outputs/output_processor.py
@@ -180,6 +180,11 @@ class OutputProcessor(LoggingConfigurable):
             if output_index is not None and output_index < len(target_cell["outputs"]):
                 target_cell["outputs"][output_index] = output
             else:
+                if output_index is not None:
+                    self.log.warning(
+                        f"Stale output index {output_index} for display_id '{display_id}' "
+                        f"(outputs length: {len(target_cell['outputs'])}), appending instead."
+                    )
                 target_cell["outputs"].append(output)
             self.log.debug(f"Wrote output to ydoc: {path} {cell_id} {output}")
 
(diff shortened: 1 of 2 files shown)


jupyter_server_documents/outputs/output_processor.py
```diff
@@ -177,7 +177,7 @@ async def output_task(self, msg_type, cell_id, content):
         _, target_cell = notebook.find_cell(cell_id)
         if target_cell is not None and output is not None:
             output_index = self.outputs_manager.get_output_index(display_id) if display_id else None
-            if output_index is not None:
+            if output_index is not None and output_index < len(target_cell["outputs"]):
                 target_cell["outputs"][output_index] = output
             else:
                 target_cell["outputs"].append(output)
```
3coins, line 188: **minor**: Adding a warning about this condition would help.

```suggestion
        if output_index is not None:                                                                                                           
              self.log.warning(                                                                                                                  
                  f"Stale output index {output_index} for display_id '{display_id}' "                                                            
                  f"(outputs length: {len(target_cell['outputs'])}), appending instead."                                                         
              )                                                                                                                                  
          target_cell["outputs"].append(output)  
```

3coins (approved):
@j1wonpark 
Thanks for submitting this fix. Overall looks good!

One minor comment about adding a warning if this out of bounds occur and we resort to appending the output.

dlqqq (approved):
Looks great, thank you for helping fix this bug! Merging.

Issue #197, opened 2026-03-20
Bugs when adding Unicode chars through Python API


dlqqq:
(likely an upstream bug in `pycrdt`, but opening here to track)

Myself & others have noticed that when agents use `jupyter_ai_tools` to add certain non-Latin Unicode characters, we are seeing errors in the console. I have also noticed that this triggers an "off-by-one" error where every character added after the Unicode character gets added to the wrong index.

```
994.bc087314bdabeecb0923.js?v=bc087314bdabeecb0923:1 Caught error while handling a Yjs update Error: Unexpected case
    at an (9046.99c477ea375dcbb8c7ca.js?v=99c477ea375dcbb8c7ca:1:17303)
    at dn (9046.99c477ea375dcbb8c7ca.js?v=99c477ea375dcbb8c7ca:1:17384)
    at Ao (9046.99c477ea375dcbb8c7ca.js?v=99c477ea375dcbb8c7ca:1:49724)
    at Io (9046.99c477ea375dcbb8c7ca.js?v=99c477ea375dcbb8c7ca:1:49843)
    at Mo (9046.99c477ea375dcbb8c7ca.js?v=99c477ea375dcbb8c7ca:1:50024)
    at ch.getMissing (9046.99c477ea375dcbb8c7ca.js?v=99c477ea375dcbb8c7ca:1:99089)
    at Si (9046.99c477ea375dcbb8c7ca.js?v=99c477ea375dcbb8c7ca:1:40961)
    at 9046.99c477ea375dcbb8c7ca.js?v=99c477ea375dcbb8c7ca:1:41615
    at Jo (9046.99c477ea375dcbb8c7ca.js?v=99c477ea375dcbb8c7ca:1:54106)
    at Ci (9046.99c477ea375dcbb8c7ca.js?v=99c477ea375dcbb8c7ca:1:41525)
```

xrl:
I have taken to telling my agent to use plain ascii only. and now my notebooks are way more stable. I'm working on a reproduction.

dlqqq:
We should keep this open until https://github.com/y-crdt/pycrdt/pull/379 is addressed.

@xrl Thanks for contributing that PR. I'm going to meet with the other contributors more regularly and work with them to see what else is pending & how we can help get it across the finish line.

PR #218, opened 2026-04-10
fix: batched catchup in YRoomUpdateBuffer.resume()

xrl:
## Summary

- Ports the batched catchup approach onto the `YRoomUpdateBuffer` class introduced in #215
- Saves the server's state vector before the handshake in `handle_sync()`
- After the handshake, computes a single batched diff via `ydoc.get_update(pre_sync_sv)` and passes it to `resume()`
- `resume()` broadcasts the batched diff instead of replaying individual queued messages

This fixes two issues for divergent client handshakes:

1. **Silent data loss**: updates queued during the divergent handshake are delivered as one clean diff covering the full gap
2. **findIndexSS crash (#197)**: individual incremental Text updates from pycrdt after multi-byte characters use offset encoding that JS yjs cannot resolve. A single batched update keeps all CRDT struct references resolvable within the same message.

## Changes

**`rooms/yroom.py`** — `handle_sync()`:
- Save `pre_sync_sv = self._ydoc.get_state()` before the handshake
- After handshake, compute `catchup = self._ydoc.get_update(pre_sync_sv)`
- Pass batched catchup to `self.update_buffer.resume(catchup_message=...)`

**`rooms/yroom_update_buffer.py`** — `resume()`:
- Accept optional `catchup_message` parameter
- Discard the queue and broadcast the batched message instead of replaying individual updates

## Test plan

- [x] All 22 tests pass (`pytest jupyter_server_documents/tests/test_yroom_sync.py jupyter_server_documents/tests/test_yroom.py`)
- [x] Stress tests in separate PR: #219

## Related

- Upstream pycrdt fix for the offset encoding bug: https://github.com/y-crdt/pycrdt/pull/379
- #197 (findIndexSS crash with Unicode chars)
- #215 (YRoomUpdateBuffer introduction)
- #219 (stress tests, split out per @dlqqq's request)

### DIFF
--- a/jupyter_server_documents/rooms/yroom.py
+++ b/jupyter_server_documents/rooms/yroom.py
@@ -677,6 +677,14 @@ class YRoom(LoggingConfigurable):
         # Mark client as desynced
         self.clients.mark_desynced(client_id)
 
+        # Save the server's state vector before the handshake. After the
+        # handshake completes, get_update(pre_sync_sv) produces a single
+        # batched diff covering any mutations that occurred during the
+        # handshake gap. This replaces replaying individual queued messages,
+        # which triggers a pycrdt offset encoding bug (#197) where
+        # incremental Text updates after multi-byte characters crash JS yjs.
+        pre_sync_sv = self._ydoc.get_state()
+
         # Check if client has divergent history
         ss1_payload = ss1_message[1:]
         divergent = self._has_divergent_history(ss1_payload)
@@ -728,13 +736,18 @@ class YRoom(LoggingConfigurable):
             self.log.exception("Exception raised during sync handshake with client '%s' in room '%s':", client_id, self.room_id)
             handshake_failed = True
 
-        # Restore the YDoc source, flush all document updates queued in the
-        # update_buffer, and resume saving, regardless of whether the handshake
+        # Restore the YDoc source, flush the update_buffer with a batched
+        # catchup diff, and resume saving, regardless of whether the handshake
         # succeeded.
         if saved_source is not None and self._jupyter_ydoc is not None:
             self._jupyter_ydoc.source = saved_source
             self.log.info("Restored YDoc source.")
-        self.update_buffer.resume()
+        catchup = self._ydoc.get_update(pre_sync_sv)
+        # An empty yjs update is 2 bytes (b"\x00\x00").
+        catchup_message = None
+        if catchup and len(catchup) > 2:
+            catchup_message = pycrdt.create_update_message(catchup)
+        self.update_buffer.resume(catchup_message=catchup_message)
         if self.file_api is not None:
             self.file_api._reloading_content = False
         
--- a/jupyter_server_documents/rooms/yroom_update_buffer.py
+++ b/jupyter_server_documents/rooms/yroom_update_buffer.py
@@ -41,13 +41,20 @@ class YRoomUpdateBuffer(LoggingConfigurable):
         """Start queuing updates instead of broadcasting them."""
         self._paused = True
 
-    def resume(self) -> None:
-        """Broadcast all queued updates and unpause."""
-        queued = self._queue
+    def resume(self, catchup_message: bytes | None = None) -> None:
+        """Discard queued updates and unpause. If catchup_message is provided,
+        broadcast it as a single batched update instead of replaying individual
+        queued messages.
+
+        Batching avoids a pycrdt offset encoding bug
+        (jupyter-ai-contrib/jupyter-server-documents#197) where individual
+        incremental Text updates after multi-byte characters crash JS yjs
+        clients with findIndexSS "Unexpected case".
+        """
         self._queue = []
         self._paused = False
-        for message in queued:
-            self._broadcast(message)
+        if catchup_message is not None:
+            self._broadcast(catchup_message)
 
     def _broadcast(self, message: bytes) -> None:
         """Send a message to all synced clients."""


dlqqq:
@xrl Thank you for opening this PR! Really appreciate your interest in this package. 🤗 

We just opened a PR that does something similar in #215. Although it was motivated by an issue different from the one you're reporting, I believe the implementation is similar. In #215 we are introducing a mechanism to make clients sync one at a time, queuing updates while a client is syncing, and then flushing the queue of updates after the client completed syncing.

However it does not implement the "batched catchup diff" approach where you merge all of the updates into a single message, which you discovered. Great find! Currently we plan to merge #215 ASAP to unblock the JSD v0.2.0 release, which will deploy the memory management feature that admins really need in prod deployments.

Could you help port over your changes to the new `YRoomUpdateBuffer` class after #215 is merged? Would love to see your stress test in a separate PR too.

xrl:
@dlqqq Thanks for the review and for merging #215!

I've rebased this PR onto main (absorbing #215) and reworked the implementation to use `YRoomUpdateBuffer`:

- **`resume()`** now accepts an optional `catchup_message` parameter. Instead of replaying individual queued messages (which triggers the pycrdt offset encoding bug from #197), it broadcasts a single batched diff computed from the pre-handshake state vector.
- **`handle_sync()`** saves `pre_sync_sv = self._ydoc.get_state()` before the handshake and passes `ydoc.get_update(pre_sync_sv)` to `resume()` after it completes.

Stress tests are in a separate PR: #219

The diff is now just 2 files changed, +28/-8 lines.

dlqqq:
Thanks for working on this & for rebasing the PR! We will definitely give this a review and merge following JSD v0.2.0 release.

jupyter_server_documents/rooms/yroom.py
```diff
@@ -677,6 +677,14 @@ async def handle_sync(self, client_id: str, ss1_message: bytes) -> None:
...
+        # handshake completes, get_update(pre_sync_sv) produces a single
+        # batched diff covering any mutations that occurred during the
+        # handshake gap. This replaces replaying individual queued messages,
+        # which triggers a pycrdt offset encoding bug (#197) where
+        # incremental Text updates after multi-byte characters crash JS yjs.
+        pre_sync_sv = self._ydoc.get_state()
+
         # Check if client has divergent history
         ss1_payload = ss1_message[1:]
         divergent = self._has_divergent_history(ss1_payload)
```
dlqqq, line 690: We are calling `self._ydoc.get_state()` twice - once here and another in `self._has_divergent_history()`.

To avoid this duplicate computation, could we have `_has_divergent_history()` accept the server state vector (in bytes) as an additional argument? e.g. `client_ss1` and `server_sv_bytes`.

jupyter_server_documents/rooms/yroom_update_buffer.py
```diff
@@ -41,13 +41,20 @@ def pause(self) -> None:
...
+        (jupyter-ai-contrib/jupyter-server-documents#197) where individual
+        incremental Text updates after multi-byte characters crash JS yjs
+        clients with findIndexSS "Unexpected case".
+        """
         self._queue = []
         self._paused = False
-        for message in queued:
-            self._broadcast(message)
+        if catchup_message is not None:
+            self._broadcast(catchup_message)
```
dlqqq, line 57: It seems like we no longer need to store the update messages at all, since we always apply incoming YDoc updates to the YDoc so they will always be included with the catchup message. Also this method does nothing if `catchup_message is None`, which is expected. Perhaps we can require the pre-sync server state vector as an argument here, and compute the catch up message in the implementation?

The new interface would look like:

```
def resume(self, pre_sync_sv: bytes) -> None:
```

Then, we should rename this class to `YRoomUpdateChannel` or something similar that doesn't imply it's a buffer, but instead a broadcast channel that can be turned on or off. Not in scope for this PR, can be done in the future.

jupyter_server_documents/rooms/yroom.py
```diff
@@ -728,13 +736,18 @@ async def handle_sync(self, client_id: str, ss1_message: bytes) -> None:
...
         if saved_source is not None and self._jupyter_ydoc is not None:
             self._jupyter_ydoc.source = saved_source
             self.log.info("Restored YDoc source.")
-        self.update_buffer.resume()
+        catchup = self._ydoc.get_update(pre_sync_sv)
+        # An empty yjs update is 2 bytes (b"\x00\x00").
+        catchup_message = None
+        if catchup and len(catchup) > 2:
+            catchup_message = pycrdt.create_update_message(catchup)
+        self.update_buffer.resume(catchup_message=catchup_message)
```
dlqqq, line 750: With the above suggestion, we can delegate this to `self.update_buffer` (which we can rename to `self.update_channel`), e.g.


```suggestion
        self.update_channel.resume(pre_sync_sv=pre_sync_sv)
```

dlqqq (approved):
Thank you for opening this PR! The catchup now happens instantly all-at-once, instead of in rapid sequential updates. This is an innovative approach. Tested it locally and verified it works well, even better than our previous experience. 🎉 

This change allows for broader simplifications in our existing code, which I documented as suggestions below. You can work on them in the same PR, or do it later in a new one. Totally up to you, let me know your thoughts before I merge (just to make sure you see them 😄).

dlqqq:
Merging this now as I haven't gotten a reply from you in a bit. Will work on the follow up comments in a separate PR. Thanks for working on this @xrl!

PR #222, opened 2026-04-30
Fix output loss from clear_output race condition

Zsailer:
When the kernel sends `clear_output` messages (used by progress bars, `IPython.display.clear_output()`, and any code that replaces output in-place), the output processor was handling each one the same way it handles a full cell re-execution: clear the YDoc output array AND delete output files from disk.

This is too heavy for an in-place output update. In cloud environments where the session/file_id lookup takes seconds, it creates a race condition — new output gets written while a stale clear is still in flight, and the late-arriving clear deletes the output that was just written. The frontend then gets 404s trying to fetch the output files.

The fix separates the two clear paths:

- **`clear_output` from kernel** (in-place output replacement): just clears the YDoc output array via `_clear_ydoc_outputs`. Cheap, no I/O.
- **`clear_cell_outputs` from execute_request** (cell re-execution): full pipeline — clears YDoc and deletes output files from disk.

No caching mechanism needed — `_clear_ydoc_outputs` already guards on `self._file_id` being set, so if no output has been processed yet it safely no-ops.

### DIFF
--- a/jupyter_server_documents/outputs/output_processor.py
+++ b/jupyter_server_documents/outputs/output_processor.py
@@ -136,14 +136,15 @@ class OutputProcessor(LoggingConfigurable):
         if wait:
             self._pending_clear_output_cells.add(cell_id)
         else:
-            await self.clear_cell_outputs(cell_id)
+            await self._clear_ydoc_outputs(cell_id)
 
     async def output_task(self, msg_type, cell_id, content):
         """A coroutine to handle output messages."""
 
         # Check for pending clear_output before processing output
         if cell_id in self._pending_clear_output_cells:
-            await self.clear_cell_outputs(cell_id)
+            await self._clear_ydoc_outputs(cell_id)
+            self._pending_clear_output_cells.discard(cell_id)
 
         # Get file_id and path from kernel session (handles renames)
         file_id, path = await self._get_file_info()
--- a/jupyter_server_documents/tests/test_output_processor.py
+++ b/jupyter_server_documents/tests/test_output_processor.py
@@ -2,7 +2,7 @@ import json
 import pytest
 from pathlib import Path
 from tempfile import TemporaryDirectory
-from unittest.mock import AsyncMock
+from unittest.mock import AsyncMock, MagicMock, patch
 from uuid import uuid4
 
 from pycrdt import Array, Map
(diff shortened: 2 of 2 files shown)


dlqqq (commented):
OK with this PR overall. Makes sense that making the `clear_output` handling faster will solve the race condition.

But long term, do we need to apply output operations one at a time to avoid race conditions? e.g. implement a message queue or locks on each method to avoid concurrent request handling?

jupyter_server_documents/outputs/output_processor.py
```diff
@@ -48,13 +48,15 @@ def yroom_manager(self):
...
-        Returns:
-            tuple: (file_id, path) if found, (None, None) otherwise
+        The file_id is looked up from the kernel session on first call
+        and cached for subsequent calls. The cache is invalidated in
+        clear_cell_outputs (called on execute_request) to handle renames.
         """
+        if self._file_id:
+            return self._file_id, None
+
         try:
```
dlqqq: I don't think this is a good change to the `_get_file_info()` call. If `self._file_id` is set, this method never returns the path. The code has to set `self._file_id = None` to get the path; `clear_cell_outputs()` does this implicitly.

IMO, we should split this into two separate methods for getting the file ID and the path. That way caching the file ID still allows paths to be retrieved.

dlqqq (changes requested):

jupyter_server_documents/outputs/output_processor.py
dlqqq: I see the problem with splitting the two. You need the path (slow) to get the file ID.

But still, I don't think whether the path is returned should be based on whether `self._file_id` is set.

jupyter_server_documents/outputs/output_processor.py
```diff
@@ -92,16 +95,16 @@ async def _clear_ydoc_outputs(self, cell_id):
...
 
     async def clear_cell_outputs(self, cell_id):
-        """Clears all outputs for a cell on disk and in ydoc."""
+        """Clears all outputs for a cell on disk and in ydoc.
 
-        # Get file_id and path from kernel session (handles renames)
+        Called on execute_request (cell re-execution). Invalidates the
+        cached file_id to handle notebook renames.
+        """
+        self._file_id = None
```
dlqqq: Why do we only invalidate the file ID when the user re-executes a cell? Can't the notebook technically be moved at any point?

Overall I understand the aim of caching the file ID to improve performance, but I'm not sure if it is safe in general to cache it and make consumers manually invalidate the cache to get a new ID.

jupyter_server_documents/outputs/output_processor.py
```diff
@@ -136,14 +139,17 @@ async def clear_output_task(self, cell_id, content):
         if wait:
             self._pending_clear_output_cells.add(cell_id)
         else:
-            await self.clear_cell_outputs(cell_id)
+            if not self._file_id:
+                await self._get_file_info()
```
dlqqq: Can we just skip this check? `self._clear_ydoc_outputs()` doesn't need the file ID or the path to be up to date.

dlqqq: If we can skip this check, then we can revert the caching of the file ID until we find a more efficient (and safe) way to do it, while also ensuring that `clear_output` messages are handled rapidly.

Zsailer: Agreed. This is exactly what I did. `clear_output_task` now just calls `_clear_ydoc_outputs` directly (which already guards on `self._file_id` being set). No `_get_file_info` call, no caching. Same fix for the race condition but much simpler.

jupyter_server_documents/outputs/output_processor.py
Zsailer: Good point. Removed the caching mechanism entirely so this concern goes away. `clear_cell_outputs` still does a fresh `_get_file_info` lookup on every call (which is fine since it only fires on execute_request), and the kernel `clear_output` path never touches `_get_file_info` at all.

jupyter_server_documents/outputs/output_processor.py
Zsailer: I've dropped the caching entirely in the updated commit.

dlqqq (approved):

dlqqq:
Seeing the same CI issue in #235. Overriding and merging - we'll tackle that before the next release.

Issue #223, opened 2026-05-04
Add kernel liveness check to `_should_free_room()` criteria


vincentye38:
# Feature Request: Add kernel liveness check to `_should_free_room()` criteria

## Background

`_auto_free_rooms()` exists to prevent `YRoom` memory leaks when notebooks are opened and closed with their kernels already shut down. Without this GC, orphaned `YRoom` objects accumulate in memory indefinitely.

The current criteria to identify a freeable room are:

- `room.empty` — no WebSocket clients connected
- `room.inactive` — no activity for `inactivity_timeout` seconds
- `execution_state in {"idle", "dead"}` — awareness state set by the kernel

## Problem

The criteria do not verify whether the kernel is actually alive. `execution_state = "idle"` is set server-side by `DocumentAwareKernelClient._handle_status_message()` whenever the kernel finishes executing — including when the browser is disconnected. This causes the following use case to fail:

A user submits a long-running Spark cell, then closes their laptop. The cell completes on the backend — the kernel sends `status: idle`, updating `execution_state` to `"idle"` server-side. With no browser clients connected, the room now satisfies all three GC conditions. `_auto_free_rooms()` frees the room, the stop callback fires `delete_session()`, and the kernel is terminated along with its Spark session. When the user reopens their laptop and reconnects, the kernel is dead and all in-memory results are lost — despite the job having completed successfully.

## Proposal

Add a kernel liveness check as an additional gate in `_should_free_room()`. A room should only be eligible for GC if no live kernel is associated with its session:

```python
def _should_free_room(self, room: YRoom) -> bool:
    if not room.room_id.startswith("json:notebook:"):
        return room.inactive_and_empty

    if not room.inactive_and_empty:
        return False

    # Only free the room if no live kernel exists for this room's session
    session_manager = self.parent.serverapp.session_manager
    kernel_manager = self.parent.serverapp.kernel_manager
    for session_id, rid in session_manager._room_ids.items():
        if rid == room.room_id:
            kernel_id = session_manager._kernel_id_from_session(session_id)
            if kernel_id and kernel_manager.is_alive(kernel_id):
                return False  # live kernel exists — not safe to GC

    return True
```

This preserves the original intent of `_auto_free_rooms()` — reclaiming memory from rooms whose kernels are genuinely gone — while preventing it from destroying kernels that are alive but temporarily have no connected browser clients.

## Key Files

- `rooms/yroom_manager.py` — `_should_free_room()`
[...]

PR #224, opened 2026-05-04
Allow GC to free notebook rooms with unknown execution state

Zsailer:
When a notebook room's kernel is shut down externally (or was never started),
the execution_state in awareness remains None or "unknown". Previously,
`_should_free_room` only allowed freeing rooms with execution_state "idle" or
"dead", so these orphaned rooms would stay in memory forever.

Now None and "unknown" are treated as freeable states. The `inactive_and_empty`
guard still prevents freeing rooms with connected clients or recent activity, so
there's no risk of dropping state that someone is actively using.

Also adds focused unit tests for `_should_free_room` covering all execution
state variants — idle, dead, None, "unknown", missing kernel key, busy,
starting, and the inactive guard.

### DIFF
--- a/jupyter_server_documents/rooms/yroom_manager.py
+++ b/jupyter_server_documents/rooms/yroom_manager.py
@@ -307,9 +307,9 @@ class YRoomManager(LoggingConfigurable):
 
         - For rooms providing notebooks: This task stops the room if it has is
         inactive, has no connected clients, and its kernel execution status is
-        'idle', 'dead', or unknown (None). An unknown state means no kernel
-        has reported status for this notebook — e.g. it was opened without a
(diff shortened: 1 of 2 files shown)


dlqqq:
@Zsailer Thanks for opening this PR. We just had a discussion with Vincent & Brian about the notebook room freeing behavior in #223, which seems to contradict the logic introduced by this PR. Can you check that out & share what you think the ideal behavior for freeing notebook rooms should be?

dlqqq (approved):
Thanks for adding documentation on what the "unknown" and `None` states mean! I had those originally but dropped them since I didn't know when that occur.

And thank you for including unit tests as well. :)

PR #227, opened 2026-05-04
fix: batched catchup in YRoomUpdateChannel.resume()

dlqqq:
Addresses review feedback from #218. The duplicate `self._ydoc.get_state()` call is eliminated by computing the server state vector once in `handle_sync()` and passing it to both `_has_divergent_history()` and `resume()`. The catchup diff computation is moved into `resume(pre_sync_sv)` so the caller no longer needs to assemble the catchup message—`YRoomUpdateChannel` owns that responsibility end-to-end.

Also renames `YRoomUpdateBuffer` to `YRoomUpdateChannel` (and the corresponding file and attribute) to better reflect that it is a broadcast channel that can be paused/resumed, not a traditional buffer that replays queued messages.

### DIFF
--- a/jupyter_server_documents/rooms/yroom.py
+++ b/jupyter_server_documents/rooms/yroom.py
@@ -14,7 +14,7 @@ import traitlets
 from ..websockets import YjsClientGroup
 from .yroom_file_api import YRoomFileAPI
 from .yroom_events_api import YRoomEventsAPI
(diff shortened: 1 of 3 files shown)


Zsailer (approved):

Zsailer:
LGTM! Thanks @dlqqq

PR #229, opened 2026-05-04
Decouple room GC from kernel shutdown

### LINKED ISSUE #223: Add kernel liveness check to `_should_free_room()` criteria

Zsailer:
Fixes #223

When a notebook room was freed by the GC, a stop callback fired
`delete_session()` which shut down the kernel via
`kernel_manager.shutdown_kernel()`. This caused users to lose kernel state
(variables, Spark sessions, etc.) if they closed their browser while the
kernel was idle — the GC would free the room and kill the kernel.

The stop callback now only removes the yroom from the kernel client,
preventing the freed room from receiving further messages. The kernel and
session stay alive. When the user reconnects, `_ensure_yroom_connected`
re-establishes the room-to-kernel link automatically.

This is a one-line change:
```python
# Before (kills kernel when room is GC'd):
yroom.add_stop_callback(lambda: self.delete_session(session_id))

# After (disconnects room, kernel stays alive):
yroom.add_stop_callback(lambda: kernel_client.remove_yroom(yroom))
```

Also adds unit tests verifying the stop callback behavior — that it
disconnects the room, does NOT shut down the kernel, and preserves the
session mapping for reconnection.

### DIFF
--- a/jupyter_server_documents/session_manager.py
+++ b/jupyter_server_documents/session_manager.py
@@ -130,7 +130,7 @@ class YDocSessionManager(SessionManager):
         # the yroom.
         kernel_client = self.get_kernel_client(kernel_id)
         await kernel_client.add_yroom(yroom)
-        yroom.add_stop_callback(lambda: self.delete_session(session_id))
+        yroom.add_stop_callback(lambda: kernel_client.remove_yroom(yroom))
         self.log.info(f"Connected yroom {yroom.room_id} to kernel {kernel_id}. yroom: {yroom}")
         return session_model
     
--- a/jupyter_server_documents/tests/test_session_manager.py
+++ b/jupyter_server_documents/tests/test_session_manager.py
@@ -1,0 +1,228 @@
+"""Tests for the YDocSessionManager stop callback behavior.
+
+When create_session() sets up a notebook, it registers a stop callback on
+the yroom. This callback fires when the room is freed by the GC (via
+yroom.stop()). The GC frees rooms that are inactive and empty (no WebSocket
+clients connected for a while).
+
+The stop callback must only disconnect the room from the kernel client — it
+must NOT delete the session or shut down the kernel. Room GC reclaims memory;
+kernel lifecycle is the user's responsibility.
+"""
+import asyncio
+import pytest
+from unittest.mock import AsyncMock, Mock, patch
+from traitlets.config import LoggingConfigurable
+from jupyter_server.services.kernels.kernelmanager import MappingKernelManager
+from jupyter_server_documents.session_manager import YDocSessionManager
+
+
+@pytest.fixture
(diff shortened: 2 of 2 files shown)


dlqqq (approved):
Awesome work!! Tested locally and verified that notebook room freeing now no longer stops the kernel session, and that the kernel session can be resumed seamlessly after the notebook is re-opened.

The new experience is so much better. I want to get this out ASAP, maybe today or tomorrow.

PR #232, opened 2026-05-07
[Bug] Fix out-of-band reload cascade after room restart

### LINKED ISSUE #2: Documentation: Yjs network protocol
## Description

This issue documents the Yjs network protocol, and tracks progress on publishing this info through our project documentation.

## Yjs network protocol

There are 2 **types** of messages[^1]:

* **Sync messages** - used to synchronize document state
* **Awareness messages** - used to synchronize ephemeral, global data like presence, cursor positions, etc.

There are 3 **subtypes** of sync messages defined by the Yjs network protocol:

* **Sync Step 1 (SS1)**: Sent by client joining the network, looking for the latest doc state
* **Sync Step 2 (SS2)**: A server reply to a Sync Step 1 message that encodes all of the document updates missing on the client.
* **Sync update (SUp)**: Messages that reflect changes in the YDoc, issued by clients and broadcast by the server after completing SS1 + SS2.

Essentially, the SS1 & SS2 handshake establishes a connection between 2 hosts with the same YDoc. After this connection is established, the 2 hosts can send each other SUp messages to automatically update each other’s document state.

### Yjs message encoding

Each Yjs message is encoded as a byte array composed of 3 components:

```
yjs_message := typeAndSubtype * payloadLength * payload
```

Each component is itself a byte array. Here, `*` denotes the byte array concatenation operator, i.e. `[1] * [255] = [1, 255]`.

* `typeAndSubtype` can be:
    * `[0, 0]` for SS1 messages,
    * `[0, 1]` for SS2 messages,
    * `[0, 2]` for SUpdate messages, and
    * `[1]` for Awareness messages (no subtype).
    * Technically, the message type & subtype can each be more than 1 byte. The spec allows each to be a 53-bit uint as a VLQ-encoded byte array, ranging from 1-8 bytes. However, since the spec doesn’t define types & subtypes with values greater than `255`, we can regard the type & subtype as always being 1 byte each.
    * `jupyter_server_ydoc` also allows `typeAndSubtype` to be `[125]` to represent chat messages, which are handled differently for reasons unknown to me.
* `payloadLength` is a 53-bit unsigned integer as a VLQ-encoded byte array. In other words, this component represents the length of the payload in 1-8 bytes.
* `payload` contains the “real contents” of the message as a UTF-8 encoded byte array. This encoding should not be done by ourselves, and we should use libraries like `yjs` and `pycrdt` to encode messages into byte arrays. The payload differs in structure for each message subtype:
    * SS1: payload is the **state vector**. This describes the client’s state and what updates it has received from other clients.
        * Computed via `Y.encodeStateVector(ydoc)` in `yjs`.
    * SS2: payload is the set of **updates** missing on the client, sent by the server in response to a previous SS1 message.
        * Computed via `ydoc.get_update(sv)` in `pycrdt`, where `sv` is the payload from the client’s SS1 message.
    * SUp: payload is an update sent by the client.
        * Dispatched by client via `yprovider.ts` in `packages/docprovider`.
        * Handled by server via `ydoc.apply_update(update)` in `pycrdt`.
    * Awareness: payload represents the entire **awareness state**.
        * ? elaborate

### Example Yjs messages

SS1 message:

```
[0, 0] * messageLength * stateVector
```

SS2 message:

```
[0, 1] * messageLength * updates
```

## Adapting this to a client-server topology

The Yjs network protocol generally assumes a peer-to-peer network topology, where there is no single source of truth about the state of a document. However, Jupyter Server requires a client-server topology, which requires a single source of truth, since only the server's document will be saved to the filesystem.

Therefore, there are unique constraints that must be followed by our implementation of the Yjs backend as a Jupyter server extension.[^2] These are:

1. Clients must send a SS1 message to initiate the connection, and the server should never send SS1 messages.
2. The server must broadcast all SUpdate messages sent by a client to all other clients connected to the server.

## References

[^1]: Protocol definition: https://github.com/yjs/y-protocols/blob/master/PROTOCOL.md

[^2]: Guidance on implementing the protocol in a client-server topology: https://github.com/yjs/y-protocols/blob/a4dd5de756f1061eaca9c1b4ebb2013e74987a11/sync.js#L13-L40

Zsailer:
After a room is GC'd and re-initialized, running a cell triggers an infinite cycle: the output processor clears cell outputs in the YDoc, which schedules a save, but the file watcher's out-of-band check fires first and detects a stale mtime mismatch (left over from the GC'd session). This triggers an in-place reload that overwrites the cleared outputs, which triggers another save, which triggers another mismatch detection, and so on. The user sees outputs never appearing, the cell timer stuck running, and cell content getting clobbered.

## Observed behavior

From server logs on an affected notebook:

**Room GC'd after inactivity:**
```
[I 2026-05-07 15:27:47.005 ServerDocsApp] Stopping YRoom 'json:notebook:1c779c04-f50d-41ec-b034-b6013ba21056'.
[I 2026-05-07 15:27:47.005 ServerDocsApp] Stopped YRoom 'json:notebook:1c779c04-f50d-41ec-b034-b6013ba21056'.
[I 2026-05-07 15:27:47.023 ServerDocsApp] Deleted YRoom 'json:notebook:1c779c04-f50d-41ec-b034-b6013ba21056'.
```

**Room re-initialized ~1hr later when user reopens notebook, immediately restarted by session_manager reconnecting the kernel:**
```
[I 2026-05-07 16:32:54.908 ServerDocsApp] Initializing room 'json:notebook:1c779c04-f50d-41ec-b034-b6013ba21056'.
[I 2026-05-07 16:32:54.912 ServerDocsApp] Loading content for room ID 'json:notebook:1c779c04-f50d-41ec-b034-b6013ba21056', found at path: 'Untitled5.ipynb'.
[I 2026-05-07 16:32:54.949 ServerDocsApp] Loaded content for room ID 'json:notebook:1c779c04-f50d-41ec-b034-b6013ba21056'.
[E 2026-05-07 16:32:55.234 ServerDocsApp] Activity in empty YRoom 'json:notebook:1c779c04-f50d-41ec-b034-b6013ba21056' from 'restart()'.
[I 2026-05-07 16:32:55.235 ServerDocsApp] Restarted FileAPI for room 'json:notebook:1c779c04-f50d-41ec-b034-b6013ba21056'.
[I 2026-05-07 16:32:55.235 ServerDocsApp] Restarted YRoom 'json:notebook:1c779c04-f50d-41ec-b034-b6013ba21056'.
```

**User runs a cell — output processor clears outputs, file watcher detects stale mtime mismatch, reloads file in-place (overwriting the YDoc mutation):**
```
[E 2026-05-07 16:33:33.206 ServerDocsApp] Activity in empty YRoom '...' from '_on_jupyter_ydoc_update()'.
{"message": "Cleared outputs for self._file_id='1c779c04-f50d-41ec-b034-b6013ba21056', cell_index=0", ...}
[W 2026-05-07 16:33:33.377 ServerDocsApp] Out-of-band file change detected. Room ID: 'json:notebook:1c779c04-f50d-41ec-b034-b6013ba21056', Last detected change: '2026-05-07 15:27:47.017984+00:00', Most recent change: '2026-05-07 16:33:33.371336+00:00'.
[I 2026-05-07 16:33:33.377 ServerDocsApp] Reloading content in-place for room ID 'json:notebook:1c779c04-f50d-41ec-b034-b6013ba21056' at path: 'Untitled5.ipynb'.
[I 2026-05-07 16:33:33.387 ServerDocsApp] Reloaded content in-place for room ID 'json:notebook:1c779c04-f50d-41ec-b034-b6013ba21056'.
```

Note `Last detected change: '2026-05-07 15:27:47.017984+00:00'` — the stale timestamp from when the room was GC'd over an hour earlier.

**Cascade repeats every ~500ms — clear, mismatch, reload, clear, mismatch, reload...:**
```
{"message": "Cleared outputs for self._file_id='1c779c04-f50d-41ec-b034-b6013ba21056', cell_index=0", ...}
[W 2026-05-07 16:33:37.952 ServerDocsApp] Out-of-band file change detected. ...
[I 2026-05-07 16:33:37.952 ServerDocsApp] Reloading content in-place ...
[I 2026-05-07 16:33:37.959 ServerDocsApp] Reloaded content in-place ...

{"message": "Cleared outputs for self._file_id='1c779c04-f50d-41ec-b034-b6013ba21056', cell_index=0", ...}
[W 2026-05-07 16:33:57.649 ServerDocsApp] Out-of-band file change detected. ...
[I 2026-05-07 16:33:57.649 ServerDocsApp] Reloading content in-place ...

{"message": "Cleared outputs for self._file_id='1c779c04-f50d-41ec-b034-b6013ba21056', cell_index=0", ...}
[W 2026-05-07 16:34:29.441 ServerDocsApp] Out-of-band file change detected. ...
[I 2026-05-07 16:34:29.441 ServerDocsApp] Reloading content in-place ...

{"message": "Cleared outputs for self._file_id='1c779c04-f50d-41ec-b034-b6013ba21056', cell_index=0", ...}
[W 2026-05-07 16:34:31.489 ServerDocsApp] Out-of-band file change detected. ...
[I 2026-05-07 16:34:31.489 ServerDocsApp] Reloading content in-place ...
```

Cell 0 cleared 5 times in one minute. Cell 3 cleared 4 times between 16:34:57 and 16:35:21.

**Client gets 404 — output wiped before it could be fetched:**
```
{"message": "404 GET /api/outputs/1c779c04-f50d-41ec-b034-b6013ba21056/527496bf-4ab5-4f12-903a-f4bdd984f91c/0.output?1778171646220 (...) 2.11ms", ...}
```

**User interrupts kernel (gives up):**
```
{"message": "Kernel interrupted: d1270be6-4ce9-4007-8ebb-6d0a4b2581e2", ...}  @ 16:33:42
{"message": "Kernel interrupted: d1270be6-4ce9-4007-8ebb-6d0a4b2581e2", ...}  @ 16:34:21
```

## Root cause

When `restart()` calls `stop(restarting=True)`, it creates a fire-and-forget save task that writes the OLD YDoc to disk. In cloud environments with async ContentsManagers, this save can complete after the new `_load_content()` has already set `_last_modified` from the file's mtime at read time. When the delayed save finishes, the file mtime is now newer than `_last_modified`, creating the initial mismatch that seeds the cascade.

## Fix

Two changes that address the root cause and harden the design:

**1. Skip save task in `stop()` when restarting** (root cause fix)

Content will be reloaded from disk immediately after restart, so saving old state is unnecessary and creates the timestamp mismatch that triggers the cascade.

**2. Save before OOB check in `_watch_file`** (design hardening)

The old loop ordering (check then save) means that if `_last_modified` ever drifts from the file mtime while a save is pending — for any reason (clock skew, eventual consistency, concurrent writes) — you get a cascade. The new ordering saves first when `_save_scheduled=True`, updates `_last_modified`, and defers the OOB check to the next quiet tick. This makes the system self-healing against any source of mtime drift.

Fix #2 without #1 would still write stale content during restarts. Fix #1 without #2 would still be vulnerable to filesystem timing issues in distributed environments. Together they eliminate both the specific trigger and the class of related failures.

### DIFF
--- a/jupyter_server_documents/rooms/yroom.py
+++ b/jupyter_server_documents/rooms/yroom.py
@@ -1125,10 +1125,13 @@ class YRoom(LoggingConfigurable):
         self._message_queue.put_nowait(None)
         
         # If this room is a document room, stop the file API and save the
-        # previous content if `immediately=False`.
+        # previous content if `immediately=False` and not restarting. When
+        # restarting, the content will be reloaded from disk, so saving the
+        # old state is unnecessary and can cause timestamp mismatches that
+        # trigger the out-of-band reload cascade.
         if self.file_api and self._jupyter_ydoc:
             self.file_api.stop()
-            if not immediately:
+            if not immediately and not restarting:
                 prev_jupyter_ydoc = self._jupyter_ydoc
                 self._save_task = asyncio.create_task(
                     self.file_api.save(prev_jupyter_ydoc)
--- a/jupyter_server_documents/rooms/yroom.py
+++ b/jupyter_server_documents/rooms/yroom.py
@@ -1060,10 +1060,10 @@ class YRoom(LoggingConfigurable):
     def stop(self, close_code: int = 1001, immediately: bool = False, restarting: bool = False) -> None:
         """
         Stops the YRoom. This method:
-         
+
         - Disconnects all clients with the given `close_code`,
         defaulting to `1001` (server shutting down) if not given.
-        
+
         - Removes all observers and stops the `_process_message_queue()`
         background task.
 
@@ -1071,6 +1071,9 @@ class YRoom(LoggingConfigurable):
         pending updates in the message queue and save the YDoc before returning.
         Otherwise, if `immediately=True`, this method will drop all pending
         updates and not save the YDoc before returning.
+
+        - If `restarting=True`, the final save is also skipped regardless of
+        `immediately`, since content will be reloaded from disk.
 
         - Clears the YDoc, Awareness, and JupyterYDoc, freeing their memory to
         the server. This deletes the YDoc history.
--- a/jupyter_server_documents/rooms/yroom_file_api.py
+++ b/jupyter_server_documents/rooms/yroom_file_api.py
@@ -431,14 +431,17 @@ class YRoomFileAPI(LoggingConfigurable):
         while True:
             try:
                 await asyncio.sleep(self._adaptive_poll_interval)
-                await self._check_file(jupyter_ydoc)
                 if self._save_scheduled:
-                    # `asyncio.shield()` prevents the save task from being
-                    # cancelled halfway and corrupting the file. We need to
-                    # store a reference to the shielded task to prevent it from
-                    # being garbage collected (see `asyncio.shield()` docs).
+                    # Save pending changes before checking for out-of-band
+                    # modifications. Checking first would detect a mismatch
+                    # between _last_modified and the file mtime (caused by a
+                    # concurrent save task from restart or filesystem timing)
+                    # and trigger a reload that overwrites the pending YDoc
+                    # mutation — creating an infinite reload cascade.
                     save_task = self.save(jupyter_ydoc)
                     await asyncio.shield(save_task)
+                else:
+                    await self._check_file(jupyter_ydoc)
             except asyncio.CancelledError:
                 break
             except Exception:
--- a/jupyter_server_documents/tests/test_yroom_file_api.py
+++ b/jupyter_server_documents/tests/test_yroom_file_api.py
@@ -1,7 +1,10 @@
 import pytest
 import pytest_asyncio
+import asyncio
 import shutil
 from pathlib import Path
+from datetime import datetime, timezone
+from unittest.mock import patch, AsyncMock, MagicMock
 import os
 from typing import Awaitable
 import pycrdt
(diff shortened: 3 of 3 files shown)


jupyter_server_documents/rooms/yroom.py
```diff
@@ -1125,10 +1128,13 @@ def stop(self, close_code: int = 1001, immediately: bool = False, restarting: bo
...
         # If this room is a document room, stop the file API and save the
-        # previous content if `immediately=False`.
+        # previous content if `immediately=False` and not restarting. When
+        # restarting, the content will be reloaded from disk, so saving the
+        # old state is unnecessary and can cause timestamp mismatches that
+        # trigger the out-of-band reload cascade.
         if self.file_api and self._jupyter_ydoc:
             self.file_api.stop()
-            if not immediately:
+            if not immediately and not restarting:
```
dlqqq, line 1137: I'm not sure if we should skip saving if restarting the room - if there are any unwritten updates that haven't been written, then that data gets lost.

I think your custom extension can run `YRoom.restart()` and then `await YRoom.until_saved` to avoid the spurious detected OOB change issue you're experiencing.

jupyter_server_documents/rooms/yroom.py
dlqqq, line 1168: See here ^

dlqqq: Long term we might just want to drop this property and the `immediately` argument in `stop()`, and instead just change our method interface:

```py
# stop immediately
def stop_immediately()

# save before stopping - you can just await this instead of awaiting until_saved
async def stop()

# save before restarting
async def restart()
```

jupyter_server_documents/rooms/yroom_file_api.py
```diff
@@ -431,14 +431,17 @@ async def _watch_file(self, jupyter_ydoc: YBaseDoc) -> None:
...
+                    # Save pending changes before checking for out-of-band
+                    # modifications. Checking first would detect a mismatch
+                    # between _last_modified and the file mtime (caused by a
+                    # concurrent save task from restart or filesystem timing)
+                    # and trigger a reload that overwrites the pending YDoc
+                    # mutation — creating an infinite reload cascade.
                     save_task = self.save(jupyter_ydoc)
                     await asyncio.shield(save_task)
+                else:
+                    await self._check_file(jupyter_ydoc)
```
dlqqq, line 444: I'm not sure this should be the new behavior. The design principle we follow is that the filesystem is the source of truth. Edits made directly to the file always take precedence over user changes, and that was what was codified in the previous implementation.

dlqqq (changes requested):
Hey Zach, I see that you're using a custom extension that calls `YRoom.restart()` - that method is never called by JSD by default anymore.

One quirk of our current API is that consumers need to call `await YRoom.until_saved` after calling `YRoom.restart()`. I would try that first to see if that fixes your issue.

Long term, I think we can improve the `YRoom` API to avoid the pitfall you ran into here - left that as a comment below. Overall, I'm not confident that the logic introduced by this  PR is safe to release. Happy to work with you more on this issue more.

dlqqq:
Closing this PR as the workaround noted in the review should suffice - opened follow-up in #241.

PR #230, opened 2026-05-07
Fix kernel consoles

dlqqq:
## Summary

Fixes #225. Kernel consoles were completely broken after installing jupyter-server-documents because:

1. The output processor intercepted and suppressed console output messages (returning `None` from `handle_document_related_message`), preventing them from reaching the frontend websocket listeners.
2. `update_session()` and `delete_session()` would crash with `LookupError` when called on non-notebook sessions that had no YRoom registered.

## Changes

**`session_manager.py`:**
- Track console session IDs explicitly via `_console_session_ids` set
- Add `_is_console_session()` helper method
- Skip YRoom operations in `create_session`, `update_session`, and `delete_session` for console sessions

**`kernel_client.py`:**
- Skip output processor interception when no YRooms are registered (i.e. kernel consoles), letting messages pass through to listeners unmodified

## Testing

All 185 existing tests pass. Added 6 new unit tests.

PR #238, opened 2026-05-13
Bypass kernel output handling if no YRooms attached

dlqqq:
## Description

- Bypass kernel output handling if no YRooms are attached (e.g. kernel consoles or Voila dashboards) - this should be a more general and reliable approach than the one introduced in #230.
- Removed remaining references to `anyio`


dlqqq:
@Zsailer's changes will override most of this - keeping this open as reference, will close later.

dlqqq:
This PR is no longer needed since @Zsailer's pending changes will attach a `DocumentAwareKernelClient` only onto notebook `YRoom` instances, effectively doing the same thing as this PR. Closing

Issue #239, opened 2026-06-09
Jupyter Server Documents v0.3.0


dlqqq:
## Context

Jupyter Server Documents (JSD) is a single extension that provides real-time server-side document handling in JupyterLab. JSD addresses several issues we have seen in the ecosystem across multiple projects over several years -- data loss, kernel status inconsistency, content duplication, and out-of-band changes not propagating to the UI, to name a few. JSD's existence was necessary for Jupyter AI v3: without JSD, agent file edits were not appearing consistently.

JSD's purpose is **not** to supersede the work that others have spent countless hours contributing. Just as Jupyter AI v3 would not have been possible without JSD, JSD would not have been possible without the preceding work done by the community.

JSD's purpose is to _unify_ the Jupyter community around a minimal and robust implementation which can be eventually upstreamed into Jupyter Server and other packages, gathering the lessons learned from several people over several years, with the ultimate goal of creating a more reliable, stable, and performant experience for users.

## Motivation 

Following the release of Jupyter AI v3.0, we've gotten lots of feedback about Jupyter Server Documents (JSD) from other key Jupyter maintainers and developers. Thank you to everyone who took the time to voice their issues at recent community calls and events like the Jupyter AI Developer Summit at London.

The reported issues all converge on a shared thread: JSD causes compatibility issues with other extensions. While Jupyter AI + JSD work well in isolation, JSD has been known to cause issues with other packages:

- `voila`: #233 
- `nb-cli`: https://github.com/jupyter-ai-contrib/nb-cli/issues/87
- `jupyter-collaboration`'s document timeline feature

After (many hours of) investigation, we found that these incompatibilities are all caused by the many custom overrides added in JSD's development process. JSD overrides how notebook outputs are stored in `YNotebook` objects, delegating output handling to a custom output service. JSD overrides the kernel layer, which was done to provide server-side document handling and fix a variety of kernel-related bugs in upstream Jupyter Server. JSD also overrides the frontend document widgets and shared models to provide a "document reset feature" previously used to handle out-of-band changes (no longer used following #202).

## Proposal

Today, we have a much better understanding of what worked well and what didn't. We've gathered the feedback shared with us, thoroughly audited our source code, and learned from other projects. To further our goal of providing a robust extension and unify the community, we plan to address all of the compatibility issues listed above in **JSD v0.3.0**. 

Specifically, in JSD v0.3.0, we plan to make the following changes:

[...]

PR #240, opened 2026-06-09
fix: restore kernel.last_activity updates for idle monitoring

joshuatowner:
NextGenMappingKernelManager.start_watching_activity() was a no-op, which prevented kernel.last_activity from being updated on IOPub messages. This broke external idle monitoring systems (e.g. SageMaker's idle shutdown extension) that read last_activity via the Jupyter session API.

Instead of opening a second ZMQ IOPub socket (which conflicts with JSD's single-client architecture), we attach a lightweight listener callback to the kernel's main_client that updates last_activity on every IOPub message.

Fixes: kernel.last_activity frozen at kernel start time, causing premature idle shutdown during active cell execution.

Before
image

After
image

without JSD
image

### DIFF
--- a/jupyter_server_documents/kernels/kernel_manager.py
+++ b/jupyter_server_documents/kernels/kernel_manager.py
@@ -105,6 +105,7 @@ class NextGenKernelManager(AsyncKernelManager):
         # Track execution state by watching all messages that come through
         # the kernel client.
         self.main_client.add_listener(self.execution_state_listener)
+        self.main_client.add_listener(self._record_activity)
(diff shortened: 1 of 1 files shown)


dlqqq (commented):
@joshuatowner Thanks for opening this PR! 

The issue you're reporting might be fixed once we remove the custom kernel layer (specifically the `KernelManager` / `MappingKernelManager`) in JSD, which we are currently planning (#239). 

I hope we can get this done by the end of the week.

cc @Zsailer : will your upcoming changes also fix this issue since it will remove the custom kernel manager classes?

dlqqq (approved):

PR #244, opened 2026-06-09
Remove outputs service (Option 1)

### LINKED ISSUE #243: Delete or disable outputs service?
The outputs service adds complexity and has been a source of race conditions and bugs, while its primary benefit (reducing YDoc memory) is not critical for most deployments.

## Option 1: Delete outputs service

Remove all outputs service code entirely — the `OutputProcessor`, `OutputsManager`, REST handlers, disk sidecar files, and the frontend placeholder-fetching logic. Inline the output-to-YDoc writing logic directly into the kernel client. Outputs would live in the YDoc as standard nbformat objects and be saved/loaded to `.ipynb` files without any indirection.

## Option 2: Disable outputs service

Keep all existing code in place but make the outputs service disabled by default via a single configurable trait (`OutputsManager.enabled`). When disabled, outputs flow directly into the YDoc and `.ipynb` without disk storage or REST calls. Users who need the memory optimization for large outputs can opt back in via configuration.

dlqqq:
## Description

Closes #243.

This PR removes the outputs service entirely. Outputs now live directly in the YDoc and are saved/loaded as standard nbformat in the `.ipynb` file.

## Motivation

The outputs service was originally designed to keep large outputs off the YDoc by storing them as *placeholder outputs* to reduce memory pressure. In practice, this layer of indirection added complexity, required a separate disk store, and caused numerous compatibility issues (as discussed in #239).

## Demo

Tested with various types of outputs:

https://github.com/user-attachments/assets/a0e6e23a-ad24-497f-b844-4cc53c96a18d

One minor bug exists however: interactive widgets don't work on the first run, but they show up if you re-run the cell. I need to check if this is an existing JSD or if it's a regression caused by this.

## Technical details

**Backend:** The `OutputProcessor` logic that routed kernel outputs to disk and then wrote placeholders into the YDoc is gone. Instead, `DocumentAwareKernelClient.handle_document_related_message` now writes outputs directly into the YDoc cell as pycrdt `Map` objects. This includes handling `clear_output(wait=True)` semantics and `display_id`-based update-in-place. The `OutputsManager` (disk read/write, placeholder creation) and its REST handlers are deleted, along with all references in `app.py`, `YRoomFileAPI`, `YRoom`, and `YRoomManager`. The `process_loaded_notebook` / `process_saving_notebook` hooks are no longer needed since the notebook file already contains real outputs.

**Frontend:** `RtcOutputAreaModel` (which fetched outputs from the REST API when it saw a placeholder URL) is replaced by the standard `OutputAreaModel`. The `_onSharedModelChanged` handler no longer checks for `metadata.url` — it just adds the output directly. The `NotebookActions.outputCleared` handler no longer calls `DELETE /api/outputs/...` since clearing the YDoc cell outputs is sufficient (the save path picks up the empty array).

**Tests:** Deleted `test_output_processor.py` and `test_outputs_manager.py`. Updated the kernel client integration tests to verify outputs are written directly to the YDoc instead of asserting on `output_processor.process_output` mock calls.


dlqqq:
See #243, we will probably not merge this, just keeping this open for my reference before I close.

PR #247, opened 2026-06-10
Disable outputs service by default

### LINKED ISSUE #243: Delete or disable outputs service?

dlqqq:
## Description

- Supersedes #244.
- Closes #243.

This PR makes the outputs service in JSD entirely optional. To achieve this, the following changes were made:

- Sets the default value of `OutputProcessor.use_outputs_service` to `False`.
- Checks the `OutputProcessor.use_outputs_service` before initializing any outputs service related classes on the server side.
    - This ensures that the trait disables the outputs service across the entire backend, instead of just a subset, which was what happened originally.

- All frontend overrides required by the outputs service have been consolidated into a separate plugin. This plugin only loads if `OutputProcessor.use_outputs_service` is true. It does this by reading from PageConfig data now set by our server extension, which reads that trait.

### Without outputs service (new default)

https://github.com/user-attachments/assets/79e45bcb-fef4-4dc8-8ea9-6da6558051cc

## With outputs service

Set by passing `--OutputProcessor.use_outputs_service=True` when starting `jupyter-lab`:

https://github.com/user-attachments/assets/20d296e4-8c99-45fa-b6ab-166de0a88916


Zsailer (approved):
This looks good to me @dlqqq.

It needs a rebase, but otherwise good to go. I can rebase #248 after you merge.

Issue #190, opened 2026-03-04
cell execution timer can not stop


vincentye38:
The cell execution timer can not stop after the execution complete and result shown up.

<img width="1443" height="662" alt="Image" src="https://github.com/user-attachments/assets/9ad39d52-070b-4f10-917b-05791116adf8" />

Without the JSD, the timer stopped after execution complete. The cell metadata has more messages.

<img width="1445" height="622" alt="Image" src="https://github.com/user-attachments/assets/e3fa6eb7-8353-4ae9-8221-d8813e097109" />

I think, since KernelClientWebsocketConnection.exclude_msg_types is configured as following to not sent iopub.status to jupyterlab, the execute-time extension may not get proper message to stop the timer. 
https://github.com/jupyter-ai-contrib/jupyter-server-documents/blob/dfa3b62d0bb7cb1f1ee3b47c94dd234e91a5e81e/jupyter_server_documents/app.py#L110
c.KernelClientWebsocketConnection.exclude_msg_types = [
                ("status", "iopub"),
                ("stream", "iopub"),
                ("display_data", "iopub"),
                ("execute_result", "iopub"),
                ("error", "iopub"),
                ("update_display_data", "iopub"),
                ("clear_output", "iopub"),
            ]

The execute-time extension code to react to messages is in this repo:
https://github.com/deshaw/jupyterlab-execute-time

krassowski:
@Zsailer is this by design?

Zsailer:
I think the jupyterlab execution timer needs a rewrite completely. It's too brittle, given that it depends on the client being connected. 

> @Zsailer is this by design?

No, this isn't intended. We can likely fix this breakage by allowing these shell messages to flow through the kernel websocket, rather than intercepting them in the server; however, for the sake of future argument, this implementation of the cell timer is likely not where we want to be when we have a server-side state of the document.

A better design, IMO, is to move this data to the notebook awareness object (or cell metadata; however, that has as long term cost in the document's ydoc size), and let the server maintain this state. Same with kernel execution state.

krassowski:
> or cell metadata

This is what users of `jupyterlab-execute-time` expect - they can see execution time of notebooks ran in the past month when re-opening them.

> let the server maintain this state

This is what `jupyter-server-nbmodel` does, cross-reference https://github.com/datalayer/jupyter-server-nbmodel/pull/29

PR #248, opened 2026-06-10
[Feature] Server-side cell execution via REST API

Zsailer:
Closes part of #239 (items 3 and 4).

## Motivation

Several independent projects have converged on the same pattern for server-side notebook execution: a REST endpoint that triggers execution, with outputs flowing back through the YDoc sync channel. This includes [jupyverse](https://github.com/jupyter-server/jupyverse/pull/191) and [jupyter-server-nbmodel](https://github.com/datalayer/jupyter-server-nbmodel) + the pycrdt-websocket + jupyter_server_ydoc stack from jupyter-collaboration (for jupyter_server).

Our previous implementation took a different route — a custom kernel client with a listener/fan-out API, a custom kernel manager, and a message routing layer — which created compatibility issues and diverged from where the rest of the ecosystem was heading.

This PR aligns our implementation with that shared pattern. The goal isn't just to fix bugs; it's to reduce the distance between what we have and what could eventually be a single unified solution in Jupyter Server core for server-side documents and execution. The less custom machinery we carry, the easier that path becomes.

## REST API

```
POST /api/kernels/{kernel_id}/execute
```

**Request**
```json
  {
    "document_id": "string",     // required — room name (e.g. "json:notebook:<file_id>")
    "cells": [                   // required — one or more cells to execute, in order
      {
        "cell_id":     "string", // required — cell ID
        "source_hash": "string"  // required — SHA-256 hex of cell source at request time
      }
    ],
    "client_id":           "string", // optional — doc.clientID
    "request_id":          "string", // optional — UUID identifying this request
    "previous_request_id": "string"  // optional — UUID of the preceding request;
                                     // server waits until that batch is fully enqueued
                                     // before enqueuing this one (FIFO guarantee)
  }
```

**Responses**
- `200 null` — accepted (fire-and-forget)
- `400` — bad request
- `408` — predecessor timed out
- `409 {"error": "source_mismatch", "cell_id": "..."}` — source diverged

Fire-and-forget. The endpoint enqueues the cell and returns immediately. Outputs stream back to all clients via the collaboration WebSocket as the kernel produces them. Matches the [jupyverse contract](https://github.com/jupyter-server/jupyverse/pull/191) so `NotebookCellServerExecutor` from jupyter-collaboration works without modification.

## Architecture

Each YRoom owns one `AsyncKernelClient` for its kernel lifetime. An asyncio Queue + worker serializes cell executions, so "Run All" is handled naturally without any locking. The `output_hook` closure in the worker writes iopub messages directly into the YDoc cell map — `execution_state`, `execution_count`, and all output types. No background watchers, no message routing tables.

One non-obvious ZMQ fix worth calling out: we instantiate `AsyncKernelClient` directly rather than via `km.client()`. The kernel manager's `client()` clones its own session, giving every client the same ZMQ DEALER identity. The kernel's ROUTER then routes `execute_reply` to the wrong socket. A fresh `AsyncKernelClient` gets an independent session and the correct reply routing.

## Complexity removed

Dropping the custom kernel machinery removes a significant surface area of overrides that were the root cause of compatibility issues with Voila, nb-cli, and others.

| Removed | Replaced by |
|---|---|
| `DocumentAwareKernelClient` + listener API | `AsyncKernelClient` owned by `YRoom` |
| `KernelMessageCache` (msg_id → cell_id mapping) | `output_hook` closure captures `ycell` directly |
| `NextGenKernelManager` + `NextGenMappingKernelManager` | Standard kernel managers — no override |
| `kernel_manager_class` / `MultiKernelManager` config overrides | Removed entirely |
| `route_iopub` dispatch layer | Inline `output_hook` in `_run_item` |
| `ExecutionManager` indirection class | Logic inlined into the HTTP handler |
| Awareness-based execution state | `execution_state` written directly to the YDoc cell map |

## WebSocket bridge

`KernelWebsocketConnection` (extends `BaseKernelWebsocketConnection`) gives each browser WebSocket its own `AsyncKernelClient`. Four asyncio tasks forward ZMQ channel messages to the browser; incoming browser messages are deserialized and sent to the appropriate ZMQ socket. No shared state between browser tabs.


Zsailer:
cc @krassowski @SylvainCorlay @vidartf @jtpio @brichet @ellisonbg @dlqqq just for visibility.

src/index.ts
```diff
@@ -85,236 +46,101 @@ export const plugin: JupyterFrontEndPlugin<void> = {
 };
 
 /**
- * Jupyter plugin creating a global awareness for RTC.
- */
-export const rtcGlobalAwarenessPlugin: JupyterFrontEndPlugin<IAwareness> = {
```
dlqqq, line 91: We should keep this because we override the one provided by Jupyter Collaboration: see #249.

src/notebook-factory/notebook-factory.ts
```diff
@@ -1,331 +1,19 @@
...
-  originalDispose.call(this);
-};
-
-CodeCellModel.ContentFactory.prototype.createOutputArea = function (
-  options: IOutputAreaModel.IOptions
-): IOutputAreaModel {
-  return new RtcOutputAreaModel(options);
+  const state: string = this.model.sharedModel.executionState;
+  const showBusy = state === 'busy' || state === 'queued';
+  this._setPrompt(showBusy ? '*' : `${this.model.executionCount || ''}`);
```
dlqqq: This can be done just by using our existing notebook factory plugin, without relying on a prototype hack:

```ts
class JsdCodeCell extends CodeCell {
  protected _updatePrompt(): void {
    const state: string = this.model.sharedModel.executionState;
    const showBusy = state === 'busy' || state === 'queued';
    this._setPrompt(showBusy ? '*' : `${this.model.executionCount || ''}`);
  }
}

export class JsdNotebookContentFactory
  extends NotebookPanel.ContentFactory
  implements NotebookPanel.IContentFactory
{
  createCodeCell(options: CodeCell.IOptions): CodeCell {
    return new JsdCodeCell(options).initializeState();
  }

  createNotebook(options: Notebook.IOptions): Notebook {
    return new ResettableNotebook(options);
  }
}
```

We'll remove `createNotebook()` in the future once we remove all of the overrides for "document resets" - fine to keep it here for now.

src/index.ts
```diff
@@ -85,236 +46,101 @@ export const plugin: JupyterFrontEndPlugin<void> = {
...
   plugin,
-  executionIndicator,
-  kernelStatus,
-  notebookFactoryPlugin,
-  codemirrorYjsPlugin,
   backupCellExecutorPlugin,
   disableSavePlugin,
-  ychat
+  notebookFactoryPlugin,
+  codemirrorYjsPlugin
```
dlqqq: I think this can be removed - it's just used to handle document resets, which we'll remove from the codebase after this PR anyways.

dlqqq (commented):
@Zsailer Great work, thank you so much!! Need to give this a full review soon, just reviewed the frontend.

I think it will be easier to merge #247 first, then address any conflicts that pop up here. Happy to help fix the merge conflicts here, I'm very familiar with the JSD lab extension now.

src/index.ts
dlqqq: (line 143 specifically)

jupyter_server_documents/rooms/yroom.py
```diff
@@ -1057,6 +1078,292 @@ def handle_inband_deletion(self) -> None:
         self.stop(close_code=4002, immediately=True)
     
 
+    # ── Kernel client lifecycle ───────────────────────────────────────────────────
```
dlqqq: Thanks for making these changes so clean! These changes are purely additive and don't seem to rely on modifying any existing methods on the `YRoom`. 

I think it makes sense to make this a new `YNotebookRoom` subclass to house all of these methods. Then `YRoomManager` just creates a `YNotebookRoom` or `YRoom` based on the room ID.

That would add type safety -- consumers should check `isinstance(room, YNotebookRoom)` before calling any methods like `connect_kernel()`, preventing silly runtime errors like trying to connect a kernel to a room serving a text file.

I can help add this change, sounds fun.

src/notebook-factory/notebook-factory.ts
krassowski: Why would any of that be even needed though? In jupyverse and jupyter-server-nbmodel there is no such code, JuypterLab proper just reads it from the YDoc state.

jupyter_server_documents/rooms/yroom.py
```diff
@@ -1057,6 +1078,292 @@ def handle_inband_deletion(self) -> None:
...
+        ycell = self._find_kernel_cell(ydoc, cell_id)
+        file_id = self.room_id.split(":", 2)[2]
+        # Clear outputs immediately on enqueue so stale results don't linger
+        # while the cell waits behind other cells in the queue.
+        if clear_outputs:
+            # Use in-place deletion, NOT ycell["outputs"] = [].
+            # Assigning a plain list would replace the pycrdt Array with a
+            # Python list, breaking YDoc sync for all connected clients.
+            del ycell["outputs"][:]
+        ycell["execution_state"] = "queued"
```
krassowski: That's interesting. Right now, the public API and frontends only know about [`idle` and `running`](https://github.com/jupyter-server/jupyter_ydoc/blob/af522e6fc657abdf95e5bcddd98dd2d574ff8b83/javascript/src/api.ts#L537). The `queued` and `running` are not being distinguished. I see you are also adding a `busy` state, which again is not currently part of the jupyter-ydoc `IExecutionState` API.

I think you are adding more private workarounds as @dlqqq highlights in https://github.com/jupyter-ai-contrib/jupyter-server-documents/pull/248#discussion_r3392832703 to support this distinction; the `JsdNotebookContentFactory` approach that David proposes would be better but still would clash with other extensions which customize the counter (e.g. adding a run button there).

I am not requesting any change, this is not my sandbox, but just want to encourage contributing back to core things that make sense (e.g. I could see value in `running` vs `queued` distinction) and staying aligned on others (e.g. I don't think using `busy` over `running` makes sense if upstream settled on `running`). The turnaround time to merge PRs in jupyter-ydoc and JupyterLab is really quick, and especially if this is to go into jupyter-server proper I don't think we should be adding yet another implementation with its own quirks but work on standardizing based on what makes the most sense.

krassowski: For context see the discussion in my PRs upstream/in `jupyter-server-nbmodel`:
- https://github.com/jupyter-server/jupyter_ydoc/pull/259
- https://github.com/jupyterlab/jupyterlab/pull/16651
- https://github.com/jupyterlab/jupyterlab/pull/17195
- https://github.com/datalayer/jupyter-server-nbmodel/pull/25

krassowski:
> what could eventually be a single unified solution in Jupyter Server core for server-side documents and execution

@davidbrochart @Zsailer 

During the jupyter-ai workshop in London I called out few problems with the existing `/execute` API, based on experience deploying `jupyter-server-nbmodel`:
- a) the YDoc notebook state is asynchronously updated; this means that user A can send a request to execute cell with given ID but in the meantime edits of user B can make its way into the YDoc state on the server; this has severe implications:
  - user's B edits could make the execution raise an error if they just started typing and have not reached a state of valid syntax
  - if user's B edits led to change in code leading to data loss (say they put in `rm -rf /` as a joke), user A may be attributed as executor of the bad code, even though he has not seen the change while he was running the execution request
  - in rare circumstances this could happen for a single user too which would be really frustrating
- b) while websockets guarantee the order of execution requests, there is no such a guarantee when we use the REST API requests. This means that quickly scheduled execution requests (e.g. when using "run all" or "restart and run all") will lead to out of order execution, which would be a critical issue for users potentially leading to data loss; this was encountered by many users in https://github.com/datalayer/jupyter-server-nbmodel/issues/65. All it takes is for small difference in latency on the router.

I proposed following two modifications:
- a) we should add source code to the payload, or a hash of the source code and compare before running; this can be added incrementally later, but in principle I believe that using only `cell_id` is not enough
- b) we should have a way to schedule running multiple cells; I initially suggested that we change the payload to `cells: []` which makes it universally robust to single-user chained execution requests. @Zsailer highlighted that a user could quickly press Shift Enter on the frontend and then the frontend also needs to be adapted to debounce/queue multiple executions before sending the request. An alternative is for the frontend to wait for a reply acknowledging that execution was scheduled, but that would lead to a severe slowdown due to network back and forth overhead; we should probably do both.


I also wanted to call out that in `jupyter-server-nbmodel` we implemented support for adding timing metadata in https://github.com/datalayer/jupyter-server-nbmodel/pull/29. This would be necessary to fix https://github.com/jupyter-ai-contrib/jupyter-server-documents/issues/190.

src/notebook-factory/notebook-factory.ts
krassowski: I answered my question in https://github.com/jupyter-ai-contrib/jupyter-server-documents/pull/248#discussion_r3394545114. If you do want to merge it with `busy` and `queued` without making a change upstream I would say (with pain) that `_setPrompt` is somewhat less annoying for downstream because it still allows other extensions to modify `CodeCell` factory.

davidbrochart:
I agree with Mike's suggestion for adding the hash of the source code to the payload, so that the server can check the current cell input against that hash and error out if it doesn't match.
For respecting cell execution order, what do you think about adding an ID to every REST execute request along with an optional "previous ID", which would be the ID of the previous execute request (if any)? The server would then be responsible for waiting for he previous execute request to be received and processed before processing the current execute request.

krassowski:
> For respecting cell execution order, what do you think about adding an ID to every REST execute request along with an optional "previous ID", which would be the ID of the previous execute request (if any)?

I did consider this, but did not suggest it because I was worried whether that would work for multiple users scheduling cell execution at the same time (where the order is not strictly defined). Would we need to synchronize the IDs? Have one set of IDs per client? 

Well since yjs gives each browser tab a unique client ID and then we can associate the client with user on jupyter-server level maybe this is a good thing to also allow tracking _who_ executed the cell?

davidbrochart:
Good point about execute requests from different users being interleaved. I guess we could pass a list of cell IDs to the execute request as you suggested then, but also adopt my suggestion above for individual cell execution requests, that would eliminate the need for debouncing in the frontend (at the cost of more state in the backend)?

Zsailer:
@krassowski @davidbrochart and @dlqqq, let me just start by expressing how grateful I am y'all are engaging and helping refine this PR! I really want to make sure we get this right. I appreciate the time you're taking to look at this.

I'll respond on the threads in line from here.

jupyter_server_documents/rooms/yroom.py
Zsailer: @krassowski Let me spend some time today re-working this logic and see if I can align with jupyter-server-nbmodel, jupyter_ydoc, and jupyterlab. Thanks for calling this out. I focused a lot on refactoring APIs in this PR and missed this detail. We should be able to achieve what we want with just `idle` and `running` today—if more states are required going forward, I will definitely contribute it upstream.

src/notebook-factory/notebook-factory.ts
Zsailer: @krassowski Let me rework this PR a bit to align with jupyter-server-nbmodel. See my comment here https://github.com/jupyter-ai-contrib/jupyter-server-documents/pull/248#discussion_r3396980053 :) 

This is left-over from our previous work. A lot of refactoring happened in this PR and I missed this once last place we are overriding. It shouldn't be necessary. :)

Zsailer:
> During the jupyter-ai workshop in London I called out few problems with the existing `/execute` API

Thanks for calling this out @krassowski! Let me take a crack at adding this to our spec here and working it into the PR. 

My plan was always to do this work too—just the scope of this PR is focused on undoing the overrides we created in this library and get aligned with others. I still think it's worth drafting up small JEP or spec doc fleshing out this API. Love your ideas here.

src/index.ts
```diff
@@ -85,236 +46,101 @@ export const plugin: JupyterFrontEndPlugin<void> = {
...
+ * cells via POST /api/kernels/{id}/execute so outputs route through the
+ * server-side YDoc rather than coming back over the kernel WebSocket.
+ *
+ * Falls back to the default WebSocket-based runCell when the flag is not set.
+ * autoStart: false means this only activates when no other implementation
+ * of INotebookCellExecutor has been provided.
  */
 export const backupCellExecutorPlugin: JupyterFrontEndPlugin<INotebookCellExecutor> =
   {
     id: '@jupyter-ai-contrib/server-documents:backup-cell-executor',
```
krassowski: Worth renaming? It's no longer backup, is it?

Zsailer:
Thanks @krassowski and @davidbrochart for the thorough feedback. In the latest commits I tried to address the source code discrepancies, execution ordering and multiple client execution attribution issue, all based on your ideas. 🚀 

Here is the spec:

`POST /api/kernels/{id}/execute`
  ```json
  {
    "cell_id":             "string",  // required — Yjs cell ID
    "document_id":         "string",  // required (or path)
    "path":                "string",  // required (or document_id)

    "source_hash":         "string",  // optional — SHA-256 hex of cell source
    "client_id":           "string",  // optional — Yjs awareness clientID
    "request_id":          "string",  // optional — UUID for this request
    "previous_request_id": "string"   // optional — UUID of preceding request
  }
```
  Responses:
```
  - 200 null — accepted (fire-and-forget)
  - 400 — bad request
  - 408 — predecessor timed out (previous request never arrived)
  - 409 {"error": "source_mismatch", "cell_id": "..."} — source diverged
```

The frontend computes SHA-256 of the cell source before sending the request. The server hashes the current YDoc source and returns 409 if they differ. This catches the case where user B edits a cell after user A presses Run — A's request is rejected rather than silently executing code A never saw. The 409 is treated as a soft failure on the frontend (a warning, not a throw) so A can decide whether to re-run with the updated source.

## Per-client request chaining

Each request carries a request_id (UUID) and an optional previous_request_id. The server waits until the predecessor has been enqueued before enqueuing the current request, guaranteeing the first-and-forget order regardless of network jitter (currently hardcoded to 10 seconds). The chain is keyed by document:client_id on the frontend, so two users' chains are completely independent — user A's ordering guarantee doesn't block user B at all.

The implementation handles the out-of-order arrival case: if B arrives before A, B creates an unset asyncio.Event for A's request_id and waits. When A arrives, it finds the existing event and sets it. If A never arrives (dropped request), B times out after 10 s with a 408. Events are cleaned up immediately after consumption so there's no memory leak.

## Attribution

client_id (the Yjs awareness clientID) is now included in every request. It's available for the server to record who executed a cell and could also be used for finer-grained per-user ordering in the future.

Let me know if anything looks off or needs further adjustment!

File: src/index.ts (hunk 1 above)
Comment (Zsailer) [reply src/index.ts:91]: :+1: Added back.

File: src/notebook-factory/notebook-factory.ts (hunk 2 above)
Comment (Zsailer) [reply src/notebook-factory/notebook-factory.ts]: @krassowski should be aligned now. No need to hack _setPrompt and no new states introduced. Just follows the same pattern as jupyverse and jupyter-server-nbmodel.

Comment (Zsailer) [reply src/notebook-factory/notebook-factory.ts]: All of our notebook overrides are now gone 🚀

File: jupyter_server_documents/rooms/yroom.py (hunk 5 above)
Comment (Zsailer) [reply jupyter_server_documents/rooms/yroom.py]: Fixed in my latest commits 🚀

File: jupyter_server_documents/rooms/yroom.py (hunk 4 above)
Comment (dlqqq) [reply jupyter_server_documents/rooms/yroom.py]: Oh, Zach already did this in latest commits, thanks!

Comment (Zsailer) [reply jupyter_server_documents/rooms/yroom.py]: yep! haha. sorry trying to update the comments as I go.

**Comment by davidbrochart** [comment]
For `POST /api/kernels/{id}/execute`:
- We should have `"cell_ids"` (not `"cell_id"`), which is an array of cell IDs (not Yjs cell IDs).
- I don't think we need the `"path"` entry at all, `"document_id"` uniquely identifies the document. Note that a document ID could be something like `type:format:path_to_document` depending on the file ID implementation.
- Should `"source_hash"` be required?
- For the `"client_id"`, should we talk about the `doc.clientID` and not mention awareness at all?

**Comment by krassowski** [comment]
> We should have "cell_ids" (not "cell_id"), which is an array of cell IDs (not Yjs cell IDs).

in that case `source_hash` would need to become `source_hashes`? It seems to me that it would be easier and to have `cells: [{}]` like:

```ts
{
   cells: [
     "cell_id": string,
     "source_hash": string
   ]
}
```

Zsailer:
@davidbrochart and @krassowski okay, sounds good. Updated the API to this:

**Request**

```
POST /api/kernels/{kernel_id}/execute
```

  ```json
  {
    "document_id": "string",     // required — Yjs room name (e.g. "json:notebook:<file_id>")
    "cells": [                   // required — one or more cells to execute, in order
      {
        "cell_id":     "string", // required — Yjs cell ID
        "source_hash": "string"  // optional — SHA-256 hex of cell source at request time
      }
    ],
    "client_id":           "string", // optional — doc.clientID from @jupyter/ydoc (attribution)
    "request_id":          "string", // optional — UUID identifying this request
    "previous_request_id": "string"  // optional — UUID of the preceding request;
                                     // server waits until that batch is fully enqueued
                                     // before enqueuing this one (FIFO guarantee)
  }
```

**Responses**

| Status | Body | When |
  |--------|------|------|
  | `200` | `null` | Cells accepted and enqueued |
  | `400` | `{"error": "..."}` | Missing field, cell not found, or cell is not a code cell |
  | `408` | `{"error": "predecessor_timeout"}` | `previous_request_id` never arrived within 10 s |
  | `409` | `{"error": "source_mismatch", "cell_id": "..."}` | Cell source changed after the user pressed Run |

File: src/index.ts [hunk 7]
```diff
@@ -86,237 +52,169 @@ export const plugin: JupyterFrontEndPlugin<void> = {
...
+          const path = sessionContext?.session?.path ?? '';
+
+          // Compute SHA-256 of the cell source so the server can detect if
+          // another user's edit arrived after this user pressed Run.
+          const source = cell.model.sharedModel.getSource();
+          const sourceBytes = new TextEncoder().encode(source);
+          const hashBuffer = await crypto.subtle.digest('SHA-256', sourceBytes);
+          const sourceHash = Array.from(new Uint8Array(hashBuffer))
+            .map(b => b.toString(16).padStart(2, '0'))
+            .join('');
```
krassowski: As per https://developer.mozilla.org/en-US/docs/Web/API/Crypto/subtle

> This feature is available only in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Defenses/Secure_Contexts) (HTTPS), in some or all [supporting browsers](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/subtle#browser_compatibility).

In JupyterLab I am avoiding using features only available in secure context because users often run on `http` outside of localhost (whether it's good or not). However, since this is not on the critical path maybe it is actually ok? For scenarios where it matters (collaboration) one could probably expect the operator to configure `https`. In any case if we go with it, these gotchas should be documented.

krassowski (commented):
> SHA-256 hex of cell source at request time

Just for context, in the debugger we already compute hashes of cells using [Murmurhash2](https://github.com/jupyterlab/jupyterlab/blob/9981272b26c8f37f25d928a1c2866673ff5fa356/packages/debugger/src/hash.ts#L22-L30). I do not remember the full rationale for why that was chosen - I think in part it was lack of availability of lightweight sha256 or similar implementation in the browsers at the time - but it still looks like potentially a better candidate because it is optimized for speed over cryptographic properties. That said, I wonder if browser-native sha256 could be faster than murmur2.

I think it is worth some benchmarking. Edit: [benchmark](https://benchjs.com/playground#/N4IgzgLghhCmIC5QDdYCcwEsD2A7RAjADQiYC2ADgDaxmy7QQ65iIDaomAJoiLAB5RKNECQBmmGriHwEIMgFc0itADoIrEgGM8cBrwEVsaCAAJTOlmbQLcpgLymAFAEoHAPlPAAOndM0zMGwlLVgACSgwAAsHUwBlCDRMXABzJxUVACYnAAM4SByiUwAGFxcAbl9zAHpqyyCaVSpsNKCQ8Mioit8AX3KQHqJOHjkAQQgAL1GAYRa0OIBVAFlsMAANABFkLgBxAEUuAE9REAkpGV5oqABaTIBWADZ1TRBLPQheWtMATWDTKKgqFMUC0oTAYFMEGwplgqDQhwgUWSKVMhz+hmMcC4pmSpjAsAgCgoz18X1GVCoqL+uFgsGxUNMXGhmAhGJMwNMYlsWiYeFM0jo2O8IBsuGFvl8bLMFjwkFMotikUOuC0zjc9k8Pj85nqgWCaFCACEEbAIY56DouOhVBbsFbcvkIDlumLtTKrP9OoaFGIxOhFQB3KCYMxaeEUKGqMAKABGEEaXEwKVNECcAHI4mFRrdHmmim0DbBjfkXeYdbK9e0ItFYqM0GgoIdVGI0NgyE4aQHTAtkhAABx1huHJwA6Le33oMpVMsz2emVRkKAUJwxjymGPqbAJJKpJwEB4uVQUKBcBJQExOTJFNPFNNTt1zmeqABW2GS6bvlTdtXq2EazVafVQmrLpen6HoAF0SBBJhUAAFSgGMAEkRhAcYplmFJ5mWVZNm2fYjhOfFCQoWYrV4XUYRVO1-UcTtTDggQIAAUWoq00FcL8qi5FVeTsDIlGySA0AQPFEmRfNaS4UTcAUMgY0nLxp0oshYmKfg7hjLgCFgABOXS7i-csPS4GAoFiW12JtNjYCcYTS38AlHLsRxTOgJp6BSREjMcswYkcfE6VMAA9ZyfICHE1K4vwAyRGhnCkUx3EcAAWNwtRnCKAGsHGnGcnDcqA2EwCDTAAMhKfhfTcAAfPKyycAqzLYABqFqSvKyrqtMAAeHrTD7Wr6vMRrCta9rSoq9Tur60x9yGh9nCa6Bxo6qaqrENxZsyNLopnHLHCcHL1t9bqACpTFUlqlsanL3HuuaD1MC6yDcE7Tq2-r5p88wcuCxw7oenaftMA7nGOrqPuey7TGuxrbqSh75uh17OumqHZu+iU3X8paYne86Ybh+GYnuzxkZet7IZmr6DzcUKsr2stEuuVKQbazAfJ6bHzDAAMQy0GInCkdL6q0SJYFMABmBBhv+ELDrGzBYdMTJJshz7HpBr5TSoXtrkTMBEJoa4aX4CBrj1ml+Wwa4xCgClEVbBQUiiMWJdV2XFpif7nCVlWCHV9HNb7bXqhhMArYtw3jdgU2mMt5JJdwW37cdqJndd938Tmr25x91zmrWyGQfMXGRzRjbCaum6K7Jx63EpyvTs23raYc7nXTLhX5frggpZ88v8ep1uXpV+G66Rp6m4J1vMYPQee9JpHDJ5+UCSUOxl88Yov07oiN4oDY4LiXgrS0Khz0llJmhjB2lL8c-L7QSXKMs9BRMY83WMtdAfKfq+nJuR8UukoLIdlEiiWEhJPEUkZJyQUmgFw8D5J-2nAAl+7o5RMDoGaOaPlejYylF4PoJwX4njoLMBg9APhyAAMSmD2AoTAWgcpnnZDsZhVpsZ0IYQAMUkJLbcCgeRKFgL4a4sCSLPFEnEQ+MorSQmhBQF+x5MGFSKGGWAMBJZRFgFQCg-oeI8mYGAIoBItCqAkTiYQtAaEwFMdUM6qhgAaCIM+MAPRRJIVsXQBgDi+RnFNO6aAyRkTrgtFERcaAsp8NdL4XhpgADqSQmCpFMIaSJ0SspgGxgQVQ8R5F2UPs8ZB2NmJwgREidJaIFAwn4EYdk-MKQROBMgYMl8YzxUiByG+2A76UjEMYGEIIYjkGoHY-xfErGuhyHMiAhxDFgDDJgCMpJw5aJ0cCXA2JiGOkZGZSUDTMRYLMIVWs9ZGzNlbO2YAzkvJRFEgQYoLzTCDGcAAfSKJgdUngflfmIZRaMqlDrnhEvyBB6A2AQV+cCesqgX5cBEbZJwUAigxlheZa6MYiilC-HMnI2NMgFJ8RMvxjBTHOAEEICZpSva+FJTQclATcDXDAIYrQmAJCql-qYFI2Bgl6JfkUKAOyqR1JpEFBkxDzLGJAQKIKaZRRphmb4AlCylkrLWWbRpoYKzylsLEVwa4Mrr0JGgOwhUEV0mRXZOSuKXSdwJb4E4esYwNiSKadgEFBggDhFgPAhAehAA):

image

src/index.ts
krassowski: Maybe beside documentation the code should be defensive (checking for support or using try-catch) if we decide that sha256 is preferred over murmur.

jupyter_server_documents/execution_handlers.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+    POST /api/kernels/{kernel_id}/execute
+
+    Server-side cell execution endpoint.
+
+    ## Request body
+
+    ```json
+    {
+      "document_id": "string",   // required — Yjs room name
```
davidbrochart: Let's not talk about Yjs in particular.
```suggestion
      "document_id": "string",   // required — room name
```

jupyter_server_documents/execution_handlers.py
```diff
@@ -0,0 +1,101 @@
...
+      "cells": [                 // required — cells to execute in order
```
davidbrochart:
```suggestion
      "cells": [                 // required — cells to execute atomically and in order
```

jupyter_server_documents/execution_handlers.py
```diff
@@ -0,0 +1,101 @@
...
+        {
+          "cell_id":     "string",  // required — Yjs cell ID
```
davidbrochart:
```suggestion
          "cell_id":     "string",  // required — cell ID
```

jupyter_server_documents/execution_handlers.py
```diff
@@ -0,0 +1,101 @@
...
+          "source_hash": "string"   // optional — SHA-256 hex of cell source
```
davidbrochart: Should it be required?

jupyter_server_documents/execution_handlers.py
```diff
@@ -0,0 +1,101 @@
...
+        }
+      ],
+
+      // Execution ordering (optional)
+      "client_id":          "string",  // doc.clientID from @jupyter/ydoc
```
davidbrochart:
```suggestion
      "client_id":          "string",  // document client ID
```

jupyter_server_documents/execution_handlers.py
krassowski: Maybe yes if using murmur hash because:
- adding it as required now and relaxing later if dowstreams ask is easier than adding as optional now and converting to required later
- it will prevent a hard to debug issue in clients

Maybe not if using sha256 as it is harder to implement in the browser.

src/index.ts
```diff
@@ -6,6 +6,63 @@ import { ISettingRegistry } from '@jupyterlab/settingregistry';
...
+    // eslint-disable-next-line no-fallthrough
+    case 1:
+      h ^= data[i] & 0xff;
+      h = (h & 0xffff) * m + ((((h >>> 16) * m) & 0xffff) << 16);
+  }
+  h ^= h >>> 13;
+  h = (h & 0xffff) * m + ((((h >>> 16) * m) & 0xffff) << 16);
+  h ^= h >>> 15;
+  return h >>> 0;
+}
```
krassowski: Maybe move it to a separate ts file to avoid confusion on what the licence extends to?

src/index.ts
Zsailer: Thanks for pointing this out, @krassowski! 

Switched to murmur2. I just copied the implementation from @jupyterlab/debugger, since it's not exported.

src/index.ts
Zsailer: Good idea. :+1: Done in the latest commit.

jupyter_server_documents/execution_handlers.py
Zsailer: Switched to Murmur2 and made it required. Added a python-side implementation too.

krassowski:
I did not dive into the code deeply yet, but I do wonder how does it work with output streams that are non-additive? There are some known issues in `jupyter-server-nbmodel` (https://github.com/datalayer/jupyter-server-nbmodel/issues/39), with example snippet being:

```python
import time
print(11111, end='\r')
time.sleep(1)
print(2, end=' ')
time.sleep(1)
print(3, end='')
print(4)
```

Producing

```
11111
2 
34
```

where `2 341` is expected. I see that `jupyter-server-documents` 0.2.x does not have this problem, but I have not tested this PR. I wonder what the difference is. From skimming the code, it appears as if there was no special logic for it so I suppose it just lets the frontend reconcile the stream and this is why it looks correct, even though it might be suboptimal on the ydoc storage level?

I am not requesting a change, but if that was a PR to jupyter-server I would suggest adding a test case to clarify the behaviour and opening an issue if this behaviour needs to be discussed.

krassowski:
Non-blocking, but could be interesting to explore in a future issue:

> 409 {"error": "source_mismatch", "cell_id": "..."}

I wonder, naively assuming that half of the mismatches would be due to the server not having processed the updates from the client, should we wait for a very short time, say 0.1s and retry before erroring out?

Zsailer:
@krassowski this PR gets the expected results. 

Screenshot 2026-06-16 at 9 16 39 AM

Might be worth discussing separate from this PR.

dlqqq (approved):
Approving for a pre-release targeting v0.3.0b0. Haven't had a chance to test yet.

dlqqq:
Great work, and thank you for this enormous contribution!

krassowski:
Note for future JEP: should we also make it work for `jupyterlab-myst` execution with user expression/markdown cells?

PR #254, opened 2026-06-13
[0.2.x] Address data loss during divergent client sync

### LINKED ISSUE #252: Data loss reported by some users in JSD v0.2.0
We have just received 2 reports of data loss from users who have upgraded to JSD v0.2. 

- We have one user on Zulip who is reporting that notebooks are being cleared and then getting suffixed with `.invalid`: https://jupyter.zulipchat.com/#narrow/channel/475130-jupyter-ai/topic/Empty.20notebooks.20suffixed.20with.20.60.2Einvalid.60/near/601238135

- We have another user who reported privately that they are also seeing data loss in `.py` files after they restart their server - these files get reset to a completely empty state.

We are taking this very seriously, although we haven't yet confirmed that these issues are caused by JSD v0.2, as both users are using very complex Python environments and deployment configurations that may be causing this issue as well.

Here are some ideas on what we can do to address this:

- Thoroughly audit changes made by #215 -- when a client closes their laptop, YRoom gets garbage collected, then reconnects, we have to clear the entire YDoc and then restore the previous content when that client finishes connecting. Although we have safety checks that should prevent JSD from saving empty content and ensure content is always restored regardless of whether the handshake succeeds, we will give this edge case another pass to ensure it's not causing any issues.

- Add tests to ensure that files don't get cleared when the server extension is stopped, in any case.

- Create backup files automatically: When the `YRoomFileAPI` detects that the file size is being reset back to 0, we will create a backup copy of a file just in case this is being caused by some bug in the `YRoom`. We will do our best to avoid creating backup files when we think the user is doing this deliberately, e.g. creating a new file, typing a line, and then deleting it. To do this we will probably use a heuristic that checks both if the new size is 0B, and if the previous size was significantly larger (e.g. >= 64B).

    - e.g. if we detect `foobar.py` gets cleared, we create a backup like `foobar_backup.py`. we will still save the empty file as we don't have a way of knowing if this change was deliberate.

    - For now we will do this by default in the interest of safety -- in the future we will make this optionally configurable as an added safety measure for admins.

- Log when this occurs: Emit a warning to the server logs in this case to allow admins to provide debugging information if they are seeing this issue occur.

We plan to make these changes ASAP and publish a 0.2 patch release containing these changes.

dlqqq:
Fixes #252.

In JSD 0.2, reconnecting to a server that had recreated its room (e.g. after idle GC) could duplicate the document's content, or worse, save an empty file to disk. The earlier fix cleared the YDoc on the server, but that left a window where a save could fire while the doc was momentarily empty.

This moves the handling to the client. During the sync handshake the client compares state vectors, and if its history has diverged from the server's — it holds client IDs the server has never seen — it clears its local doc and adopts the server's state in one transaction. Content duplication is avoided and the server never writes an empty file. The server now just logs divergent reconnects instead of acting on them.

Justification: See #257 for the complete details on how and why this works in general.

Tradeoff: on a divergent reconnect the document rebuilds from scratch, so notebook cells get removed and re-added and scroll position resets.

Tests: added Playwright E2E covering both the divergent case (no duplication, no loss) and an ordinary reconnect (offline edits preserved), for files and notebooks; see `ui-tests/AGENTS.md`. Also added E2E tests covering both `jupyterlab-chat` and `jupyter-ai-router`, ensuring that data loss, content duplication, and agent re-triggering are not caused at sync, regardless of whether the histories are divergent.

Targets `0.2.x`.

---

Note: this PR targets `0.2.x`. We still need to open the same fix against `main` so 0.3 inherits it — tracked in #258.


dlqqq:
Next steps:

- ~~Client detection of diverging server history - verify this mathematically~~ Done. See #257.
- ~~Only do this if histories are divergent~~: Done.
- ~~Fix kernels not showing up in notebook after sync~~: Done.

joshuatowner (approved):

PR #255, opened 2026-06-14
fix: remove server-side YDoc source clearing to prevent data loss

joshuatowner:
The divergent-client handshake introduced in #215 temporarily clears the YDoc source on the server during sync. If a save fires during that window (graceful shutdown or periodic _watch_file tick), it writes empty content to disk — causing data loss (V2245873628).

This commit removes the server-side source clearing entirely. Divergent clients (stale reconnections after YRoom GC) will see content duplication, which is the pre-#215 behavior — a cosmetic issue, not destructive.

Also adds a defensive guard in save() to skip writes when _reloading_content is True, as defense-in-depth against any future re-introduction of source-clearing logic.


dlqqq:
Superseded by #254.

PR #256, opened 2026-06-15
[Bug] Fix execution indicator showing 'unknown' on notebook reopen mid-execution

Zsailer:
Builds on #248 (server-side execution).

This is a recreation and simplification of what exists in Jupyter Server today: https://github.com/jupyter-server/jupyter_server/blob/9ba7a60aea07ac84307f329bdf79e70bfb3bbe56/jupyter_server/services/kernels/connection/channels.py#L158

## The Problem 

When a user refreshes the page or reopens a notebook while a long-running cell is executing, the execution indicator in the top-right shows 'unknown' instead of 'busy'.

The root cause is that new WebSocket subscribers miss the `status:busy` iopub message the kernel already sent — ZMQ PUB/SUB has no replay for late joiners.

## The fix

The fix probes kernel state on connect by sending `kernel_info_request` on both the shell and control channels simultaneously, before starting the listen tasks to avoid racing them for messages. Control bypasses the execution queue and always replies; shell is blocked while executing. This is the same "nudge" approach used in `jupyter_server`.

The inferred state (idle/busy/unknown) is delivered to the browser as a synthetic `status` iopub message so the execution indicator shows the correct state immediately on reconnect.

There's also a fast path for ipykernel >= 7.2.0 (JEP [65](https://github.com/jupyter/enhancement-proposals/blob/master/65-jupyter-xpub/jupyter-xpub.md)): listening for `iopub_welcome` on the XPUB socket, which arrives ~5ms after connection and triggers early completion instead of waiting the full window.

The nudge logic lives in `nudge.py` as a standalone function so it can be reused from any context with an `AsyncKernelClient`, not just the WebSocket handler. The two timeouts (`kernel_info_timeout`, `kernel_info_reply_window`) are configurable traitlets.

## Follow up discussion

We could remove the synthetic status message by routing these messages into the ydoc, storing them in e.g. the global/document awareness somewhere, and letting the CRDT sync the state with all clients, instead of depending on a synthetic message. That should be done in a follow up PR though.

### DIFF
--- a/jupyter_server_documents/nudge.py
+++ b/jupyter_server_documents/nudge.py
@@ -1,0 +1,152 @@
+"""
+Kernel state probing via dual-channel kernel_info_request.
+
+The nudge determines whether a running kernel is idle or busy by sending
+kernel_info_request on both the shell and control channels simultaneously.
+Control bypasses the execution queue and always replies; shell is blocked
+while a cell is executing.
+
+  control + shell reply  → "idle"
+  control reply only     → "busy"
+  no reply within timeout → "unknown"
+
+Also listens for iopub_welcome (JEP #65 / ipykernel >= 7.2.0), which arrives
+~5 ms after connection on XPUB-capable kernels and allows early termination
+of the wait before the shell_window expires.
+
+This follows the same approach used in jupyter_server for detecting kernel
+state on reconnect, generalised so it can be called from any context that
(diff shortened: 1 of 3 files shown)


dlqqq:
@Zsailer This PR still planned for merge eventually? Or is it not necessary anymore?

Issue #250, opened 2026-06-11
Execution in Console created for notebook does not show output


krassowski:
Follow steps in https://github.com/jupyterlab/jupyterlab/issues/19010

### jupyter-server-documents - no outputs show up

https://github.com/user-attachments/assets/2e9e2b30-a752-4059-8c5e-94d00ddad6c0

### jupyter-server-nbmodel - outputs show up

https://github.com/user-attachments/assets/9da56730-9208-48b4-86c9-1bd4d5a16db3

### plain JupyterLab - outputs show up

https://github.com/user-attachments/assets/59107fd5-e7e9-4dba-8fcc-7c4884d65093

# Test it yourself

https://gist.github.com/krassowski/229c9887624f8663e02962def6272ee9
https://gist.github.com/krassowski/1b683e887179ff8436986557c646ff1e

dlqqq:
Thanks, I think JSD v0.3.0 will fix this issue, but we should verify this after everything is merged. Adding it to v0.3.0 milestone.

dlqqq:
Yup, confirmed that JSD v0.3.0 will fix this issue, added E2E test here in #263. Also verified it manually.

We're still debating whether to block JSD v0.3.0 until we support Jupyter YDoc v4 & JupyterLab v4.6.0.

PR #263, opened 2026-07-01
Add E2E test for "New Console for Notebook" output rendering

dlqqq:
## Summary

Adds a Playwright/Galata UI test that captures the bug reported in #250 (and
its upstream counterpart jupyterlab/jupyterlab#19010): a code console opened via
**New Console for Notebook** rendered **no output** when code was executed in it
under `jupyter-server-documents`, while plain JupyterLab and
`jupyter-server-nbmodel` rendered it correctly.

The test asserts the **correct** behavior, so it acts as a regression guard: it
fails on a build that still has the bug and passes once the console output path
works. On the current branch the console path already works (the output
processor no longer intercepts console messages), so the test passes — and a
negative-control run confirmed it genuinely fails if the console produces no
output.

## Test scenario (`ui-tests/tests/console-for-notebook.spec.ts`)

1. Open a notebook with a live `python3` kernel, run `1 + 1` in the first cell,
   and assert the cell output is `2` (the notebook execution path works).
2. Create a console for that notebook via the `notebook:create-console` command
   (the "New Console for Notebook" menu item), sharing the notebook's kernel.
3. Run `1 + 1` in the console and assert the `2` output renders.

## Changes

- **`ui-tests/tests/console-for-notebook.spec.ts`** — the new spec. Disables
  galata's kernel/session/terminal API mocking (`kernels`/`sessions`/`terminals`
  set to `null`), since — unlike the other kernel-less specs — this one spins up
  a real kernel and galata's session-tracking route handler otherwise crashes on
  the live session-creation response under JSD.
- **`ui-tests/tests/helpers.ts`** — new shared helpers:
  - `acceptKernelDialog(page, kernel='python3')` — accept the "Select Kernel"
    dialog with a real kernel (counterpart to the existing
    `dismissKernelDialogIfPresent`).
  - `createConsoleForNotebook(page)` — run `notebook:create-console` and wait
    until the console's session has a **connected** kernel before returning
    (the prompt cell mounts before the kernel is wired up; executing too early
    silently drops the request — a startup race that flaked under parallel load).
  - `runInConsole(page, code)` — type into the console prompt with real
    keystrokes, confirm the editor holds the text, then execute via
    `console:run-forced` (avoids a focus race that a raw `Shift+Enter` keypress
    hit under parallel load).
  - `consoleOutputCount(page, text)` — count executed console cells whose
    rendered output contains `text` (excludes the prompt cell).

## Verification

- New test: **15×** parallel repeats green, plus 6× single-worker green.
- Full `ui-tests` suite: 6 passed, 4 skipped (optional chat/router deps).
- Negative control (asserting a token that can never appear) fails with the
  "issue #250" message, proving the assertion has real discriminating power.
- `prettier --check` and `eslint` clean.

Refs #250, jupyterlab/jupyterlab#19010

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
         uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
 
       - name: Install dependencies
-        run: python -m pip install -U "jupyterlab>=4.0.0,<5"
+        run: python -m pip install -U "jupyterlab>=4.0.0,<4.6"
 
       - name: Lint the extension
         run: |
@@ -82,7 +82,7 @@ jobs:
           sudo rm -rf $(which node)
           sudo rm -rf $(which node)
 
-          pip install "jupyterlab>=4.0.0,<5" jupyter_server_documents*.whl
+          pip install "jupyterlab>=4.0.0,<4.6" jupyter_server_documents*.whl
 
 
           jupyter server extension list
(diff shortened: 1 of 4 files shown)


dlqqq:
While writing this test I hit a provisioning snag worth flagging: the test suite (this new spec **and** the existing `sync.spec.ts`) fails to boot JupyterLab on **v4.6**, because 4.6 bundles `@jupyter/ydoc` v4 while JSD's frontend is built against ydoc v3. I had to pin `jupyterlab <4.6` locally to run the E2E suite.

I've opened #265 to track that separately (it's the downstream user-facing symptom of #262 — 'Upgrade Jupyter YDoc v4'). CI currently resolves `jupyterlab>=4.0.0,<5`, which will start pulling 4.6 and break the integration-tests job until #262 lands.

PR #280, opened 2026-07-20
Send current awareness state to clients on connect

### LINKED ISSUE #279: Awareness state is not sent to a client on connect (only via later change deltas)
## Summary

A newly-connected (or refreshed) client is never sent the room's **current awareness state**. It only learns existing awareness slots when a later *change delta* happens to re-touch them, so awareness that was published before the client connected can take several seconds to appear on that client.

This surfaced downstream in `jupyter-ai-persona-manager` as personas taking up to ~10s to load on page refresh (jupyter-ai-contrib/jupyter-ai-persona-manager#76): the persona list is published into awareness at startup under a fixed client ID, and a refreshed browser doesn't receive it until the next heartbeat renewal re-broadcasts that slot.

## Root cause

Awareness is broadcast **only as deltas of changed clients**:

```python
# jupyter_server_documents/rooms/yroom.py  (_on_awareness_update)
updated_clients = [v for value in changes[0].values() for v in value]
state = self._awareness.encode_awareness_update(updated_clients)
message = pycrdt.create_awareness_message(state)
self._broadcast_message(message, "AwarenessUpdate")
```

The sync handshake (`handle_sync_step1`) sends SyncStep2/SyncStep1 for the **YDoc only** and marks the client synced — but nothing sends the client the current **awareness** map. The awareness protocol has no state-vector/handshake of its own, so a fresh client cannot request it; by convention the server must push a one-time snapshot on connect. This is what `y-websocket`'s `setupWSConnection` does right after wiring up a connection:

```js
const awarenessStates = doc.awareness.getStates()
if (awarenessStates.size > 0) {
  // encodeAwarenessUpdate over ALL current client IDs, sent once to the new conn
  send(doc, conn, encodeAwarenessUpdate(doc.awareness, Array.from(awarenessStates.keys())))
}
```

Since jupyter-server-documents rolls its own room/WS layer, that snapshot-on-connect responsibility is entirely on this package's code, and it appears to be missing.

## Symptoms / how to reproduce

- Open a chat with `jupyter-ai-persona-manager` so personas are fully initialized server-side.
- Refresh the page (single tab, no other activity). Personas take up to ~10s to appear even though the data is already in the server's awareness map.
- Telling clue: with two tabs open, **closing tab B causes the persona list to load in tab A** — B's disconnect injects an extra awareness broadcast that finally reaches A. This confirms it's a delivery gap, not slow initialization.

The variable delay corresponds to waiting for the next awareness delta to re-encode the pre-existing slot, driven by heartbeat renewals (pycrdt renews once `outdated_timeout/2` = 15s has elapsed; persona-manager's own heartbeat is ~24s).

## Proposed fix

Broadcast the full current awareness state to a client **immediately after it becomes synced** — i.e. right after `mark_synced` in `handle_sync_step1` — encoding all current client IDs via `encode_awareness_update(list(self._awareness.states.keys()))` and writing it only to that client's socket (not a room-wide re-broadcast).

## Testing

- **Integration test** (`jupyter_server_documents/tests/test_yroom_sync.py`, using the existing `FakeWebSocket` + `_complete_handshake` harness): pre-populate a room awareness slot before the client connects, run the handshake, and assert the client receives an `AwarenessUpdate` carrying that slot **without any further mutation/heartbeat**. Should fail on `main`, pass after the fix. Add assertions that the snapshot goes only to the newly-synced client (not re-broadcast to existing peers) and is sent after `mark_synced` (desynced clients don't receive it).
- **E2E** for the user-visible symptom lives downstream in `jupyter-ai-persona-manager` (see that repo's #77): a solo page-refresh must repopulate the persona picker within a tight budget with no other activity.

Downstream issue: jupyter-ai-contrib/jupyter-ai-persona-manager#76

dlqqq:
Fixes #279

## Summary

A newly-connected or refreshed client is now sent the room's current awareness state as part of the sync handshake, instead of only learning about existing slots when a later change delta happens to re-touch them.

Awareness was previously broadcast solely as change deltas. A client that connected after a slot was published (e.g. a persona registering its awareness) would not see that slot until something mutated it again, which could take several seconds. This surfaces downstream in jupyter-ai-persona-manager as personas taking ~10s to appear on refresh (jupyter-ai-contrib/jupyter-ai-persona-manager#76).

## Technical details

In the SS1 handler, right after the SS2 reply (which already carries the current YDoc state) and before the SS1 request, the room encodes the full current awareness state and writes it to that client's socket alone. Grouping it with the SS2 reply means both "here's the server's current state" messages travel together and awareness reaches the client before we ask it for its own updates. It is not re-broadcast to peers that already hold the state, and it is sent only after mark_synced, so a desynced client mid-handshake never receives it.

## Testing

Backend integration tests in `test_yroom_sync.py` cover the FakeWebSocket handshake: a slot published before connect is delivered by the handshake alone, the snapshot reaches only the new client (not existing peers), and it arrives after the SS2 reply.

An end-to-end test (`ui-tests/awareness-on-connect.spec.ts`) exercises the real browser connect path that users hit on refresh: open a document, close it (the client disconnects but the room survives GC), publish a persona's awareness into the still-alive room while nothing is connected, then reopen — a fresh sync handshake, like a refresh. It asserts the persona appears in the reconnected client's awareness within 5s. This times out on `main` and passes with the fix. Backed by a test-only `/jsd-test/seed-awareness` endpoint.

### DIFF
--- a/ui-tests/jsd_test_ext.py
+++ b/ui-tests/jsd_test_ext.py
@@ -167,6 +167,45 @@ class _RecreateRoomHandler(APIHandler):
         self.finish(json.dumps({"freed": freed}))
 
 
+class _SeedAwarenessHandler(APIHandler):
+    """Publish an awareness slot into a file's room from a simulated *peer*.
+
+    This mimics another participant (e.g. an AI persona) announcing its
+    presence via awareness. The state is applied through the room's normal
+    ``handle_awareness_update`` path, so it lands in the room's awareness map
+    exactly as a real peer's would -- letting the awareness-on-connect E2E seed
+    a slot while the browser is offline, then assert the browser receives it on
+    reconnect (jupyter-ai-contrib/jupyter-server-documents#279).
+
+    Returns the peer's awareness client id (as a string) so the test can target
+    that specific slot.
+    """
+
+    @tornado.web.authenticated
+    async def post(self):
+        import pycrdt
+
+        path = self.get_argument("path")
+        token = self.get_argument("token")
+        rooms = _rooms_for_path(self.settings, path)
+        if not rooms:
+            raise tornado.web.HTTPError(404, f"No open room for path '{path}'.")
+        room = rooms[0]
+
+        # Build an awareness update from a throwaway peer and apply it to the
+        # room. The peer never renews its clock, so no later delta re-touches
+        # this slot -- which is precisely why a client that connected afterwards
+        # would never learn it without the snapshot-on-connect fix.
+        peer = pycrdt.Awareness(pycrdt.Doc())
+        peer.set_local_state_field("persona", {"name": token, "id": token})
+        update = peer.encode_awareness_update([peer.client_id])
+        room.handle_awareness_update("seed-peer", pycrdt.create_awareness_message(update))
+
+        self.finish(
+            json.dumps({"room_id": room.room_id, "peer_client_id": str(peer.client_id)})
+        )
+
+
 def _jupyter_server_extension_points():
     return [{"module": "jsd_test_ext"}]
 
@@ -185,6 +224,10 @@ def _load_jupyter_server_extension(server_app):
             (
                 url_path_join(base_url, "jsd-test", "router-fires"),
                 _RouterFiresHandler,
+            ),
+            (
+                url_path_join(base_url, "jsd-test", "seed-awareness"),
+                _SeedAwarenessHandler,
             ),
         ],
     )
--- a/jupyter_server_documents/rooms/yroom.py
+++ b/jupyter_server_documents/rooms/yroom.py
@@ -759,6 +759,28 @@ class YRoom(LoggingConfigurable):
             )
             raise
 
+        # Send the room's *current* awareness state to this client only.
+        #
+        # Awareness is otherwise broadcast solely as change deltas (see
+        # `_on_awareness_update`), so a newly-synced client would not learn any
+        # slot published before it connected until a later delta happened to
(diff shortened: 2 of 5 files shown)


dlqqq:
Haven't verified that this fixes the upstream issue, need to test.

dlqqq:
Does fix this issue


https://github.com/user-attachments/assets/ecb5f5a3-edf6-42aa-8762-7906fecf528d

ui-tests/jsd_test_ext.py
```diff
@@ -167,6 +167,45 @@ async def post(self):
...
+    ``handle_awareness_update`` path, so it lands in the room's awareness map
+    exactly as a real peer's would -- letting the awareness-on-connect E2E seed
+    a slot while the browser is offline, then assert the browser receives it on
+    reconnect (jupyter-ai-contrib/jupyter-server-documents#279).
+
+    Returns the peer's awareness client id (as a string) so the test can target
+    that specific slot.
+    """
+
+    @tornado.web.authenticated
```
krassowski, line 184: Non-blocking: should it also be authorizing? I am not sure, just noting that other thigs have both decorators.

dlqqq: Thanks for the review! This is a handler defined by a server extension only used in our E2E test setup, I think it's fine.

dlqqq:
Going to publish this as JSD v0.3.1 now since it's a release blocker for Jupyter AI v3.1.0.

PR #285, opened 2026-07-29
Follow Jupyter Server conventions on Same-Origin and auth checks, fix reverse proxy deployments

bsundaram1:
## Problem

`YRoomWebsocket` subclasses bare Tornado `WebSocketHandler` and never overrides `check_origin`. Tornado's default enforces strict same-origin (`Origin` must equal `Host`), so **every** room websocket (`JupyterLab:globalAwareness`, every `json:notebook:*`, chat rooms) gets rejected with a 403 at the handshake behind any reverse proxy that doesn't make the browser's `Origin` match what the backend sees as `Host`. 
This breaks notebook rendering and chat entirely in proxied deployments (`ServerApp.allow_origin` doesn't help since that's only consulted by `JupyterHandler`-based handlers).

## Fix

Mirrors `jupyter_server_ydoc.YDocWebSocketHandler`, which solved this same problem previously: 
inherit `JupyterHandler`, defer to authentication instead of origin-checking (`check_origin` always returns `True`), and gate access via `current_user` in `get()`. 
`prepare()` now awaits `super().prepare()` so `current_user` is populated by `JupyterHandler`'s auth resolution before `get()` checks it.

## Testing

- Added a unit test asserting `check_origin` returns `True` for a mismatched origin.
- Manually verified against a reverse proxy that intentionally mismatches `Origin`/`Host`: valid token now succeeds (previously 403), invalid/missing tokens are still correctly rejected with 403.

### DIFF
--- a/jupyter_server_documents/websockets/yroom_ws.py
+++ b/jupyter_server_documents/websockets/yroom_ws.py
@@ -51,10 +51,8 @@ class YRoomWebsocket(WebSocketHandler, JupyterHandler):
         # and skips the check for token-authenticated requests.
         return JupyterHandler.check_origin(self, origin)
 
+    @ws_authenticated
     async def get(self, *args, **kwargs):
-        if self.current_user is None:
-            self.log.warning("Couldn't authenticate WebSocket connection")
-            raise web.HTTPError(403)
         return await super().get(*args, **kwargs)
 
 
--- a/jupyter_server_documents/websockets/yroom_ws.py
+++ b/jupyter_server_documents/websockets/yroom_ws.py
@@ -1,12 +1,12 @@
 from __future__ import annotations
 from tornado.httpclient import HTTPError
 from tornado.websocket import WebSocketHandler
-from tornado import web
 from typing import TYPE_CHECKING
 import asyncio
 from ..rooms import YRoomManager
 import logging
 from jupyter_server.base.handlers import JupyterHandler
+from jupyter_server.auth.decorator import ws_authenticated
 
 if TYPE_CHECKING:
     from jupyter_server_fileid.manager import BaseFileIdManager
--- a/jupyter_server_documents/websockets/yroom_ws.py
+++ b/jupyter_server_documents/websockets/yroom_ws.py
@@ -59,7 +59,7 @@ class YRoomWebsocket(WebSocketHandler, JupyterHandler):
 
 
     async def prepare(self):
-        await super().prepare()
+        await super().prepare(_redirect_to_login=False)
 
         # Bind `room_id` attribute
         request_path: str = self.request.path
--- a/jupyter_server_documents/websockets/yroom_ws.py
+++ b/jupyter_server_documents/websockets/yroom_ws.py
@@ -45,8 +45,11 @@ class YRoomWebsocket(WebSocketHandler, JupyterHandler):
 
 
     def check_origin(self, origin):
-        # Mirroring jupyter_server_ydoc.YDocWebSocketHandler.
-        return True
+        # `WebSocketHandler` precedes `JupyterHandler` in the MRO, so `super()`
+        # would resolve to Tornado's strict same-origin check. Call
+        # `JupyterHandler`'s explicitly: it honors `allow_origin`/`allow_origin_pat`
+        # and skips the check for token-authenticated requests.
+        return JupyterHandler.check_origin(self, origin)
 
     async def get(self, *args, **kwargs):
         if self.current_user is None:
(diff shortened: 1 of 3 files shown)


jupyter_server_documents/websockets/yroom_ws.py
```diff
@@ -42,7 +44,20 @@ def contents_manager(self) -> AsyncContentsManager | ContentsManager:
         return self.settings["contents_manager"]
 
 
-    def prepare(self):
+    def check_origin(self, origin):
+        # Mirroring jupyter_server_ydoc.YDocWebSocketHandler.
+        return True
```
dlqqq: While this is faithful to how `jupyter_collaboration` does it, I'm wondering if there is a security gap introduced by this as-is.

Some services vendor JupyterLab in an `iframe`, and this requires the server to set `SameSite=None` in its `Set-Cookie` response header, i.e. cross site requests can include cookies. Previously, cross-origin attacks were blocked even with `SameSite=None` since the original `WebSocketHandler.check_origin()` method would block this.

However, now that this method always returns `True`, those requests would be allowed. That would introduce a security vulnerability for those services.

The right thing to do, IMO, is to use `JupyterHandler.check_origin()` here. It still enforces the Same-Origin check by default, but it also respects configurable traits that allow you to allowlist your reverse proxy domain in your deployments.

```suggestion
    def check_origin(self, origin):
        # `WebSocketHandler` precedes `JupyterHandler` in the MRO, so `super()`
        # would resolve to Tornado's strict same-origin check. Call
        # `JupyterHandler`'s explicitly: it honors `allow_origin`/`allow_origin_pat`
        # and skips the check for token-authenticated requests.
        return JupyterHandler.check_origin(self, origin)
```

This tighter security does require reverse proxies to either:

- Forward the client host (`proxy_set_header Host $host;`),

- Configure `ServerApp.allow_origin` or `ServerApp.allow_origin_pat` to allow the browser's origin, or
 
- Inject the token at the proxy (proxy_set_header Authorization "token $JUPYTER_TOKEN";). This is only safe if the proxy already authenticates users itself.

Your reverse proxy probably already does one of these, since most of the Jupyter APIs (including Kernel websockets) would not work without this configuration already being set.

jupyter_server_documents/websockets/yroom_ws.py
```diff
@@ -42,7 +44,20 @@ def contents_manager(self) -> AsyncContentsManager | ContentsManager:
...
+
+    async def get(self, *args, **kwargs):
+        if self.current_user is None:
+            self.log.warning("Couldn't authenticate WebSocket connection")
+            raise web.HTTPError(403)
+        return await super().get(*args, **kwargs)
+
+
+    async def prepare(self):
+        await super().prepare()
```
dlqqq: To stop redirects on unauthenticated WebSocket requests, this flag also has to get passed to `JupyterHandler.prepare()`:

```suggestion
    async def prepare(self):
        await super().prepare(_redirect_to_login=False)
```

Reference: https://github.com/jupyter-server/jupyter_server/blob/9a4d6eea2b16815a493b11fffe0b51b1fe55a81b/jupyter_server/base/handlers.py#L645-L656

jupyter_server_documents/websockets/yroom_ws.py
```diff
@@ -42,7 +44,20 @@ def contents_manager(self) -> AsyncContentsManager | ContentsManager:
...
-    def prepare(self):
+    def check_origin(self, origin):
+        # Mirroring jupyter_server_ydoc.YDocWebSocketHandler.
+        return True
+
+    async def get(self, *args, **kwargs):
+        if self.current_user is None:
+            self.log.warning("Couldn't authenticate WebSocket connection")
+            raise web.HTTPError(403)
+        return await super().get(*args, **kwargs)
```
dlqqq, line 56: This is handled by the [`@ws_authenticated`](https://github.com/jupyter-server/jupyter_server/blob/9a4d6eea2b16815a493b11fffe0b51b1fe55a81b/jupyter_server/auth/decorator.py#L118-L143) decorator provided by `jupyter_server.auth.decorator`, which does a couple of other extra things that are useful.

With that, this can be simplified to:

```suggestion
    @ws_authenticated
    async def get(self, *args, **kwargs):
        return await super().get(*args, **kwargs)
```

Note that this requires bumping to `jupyter_server>=2.13`.

dlqqq (changes requested):
@bsundaram1 Thank you for opening this PR! I like how this strengthens the YRoom WebSocket security and mirrors how Jupyter Collaboration behaves.

I have a few extra suggestions on how to strengthen the `check_origin()` method, as well as a few other minor suggestions. I can help take this over if you're short on time.

@Zsailer Can you double-check my review comments? You're more familiar with Jupyter Server than I am.

bsundaram1:
@dlqqq Thanks for the review! I've accepted your proposed changes - they make sense. I've also added a couple of new unit tests for the cases you mentioned.

dlqqq (approved):
@bsundaram1 Thanks for the fast response! This will get included in the next Jupyter AI patch. :)

PR #292, opened 2026-08-06
Skip redundant saves when rooms are stopped

dlqqq:
## Motivation

If a server is forcibly stopped while it is shutting down, active files can be truncated to 0 bytes, generating a `.invalid` sidecar file as well. This is because "pulling out the power cord" in the middle of a file write is not guaranteed to be atomically safe on all systems. We are still actively investigating exactly what scenarios allow for this.

One factor that contributes to the impact of this issue is that `YRoom.stop()` always performs a final save even when autosave has already persisted everything.

## Summary

Adds two stopgap protections around saving on stop:

1. The final save on a graceful stop now runs only when the document has unsaved changes. Edits made just before the stop, and updates applied while draining the message queue during stop, are still saved; the common already-persisted case is skipped.

2. A notebook save is skipped (with a warning) when its content is empty (0B). This should never happen, but this is being included regardless as an additional safety measure.

The save-on-stop decision (saved vs skipped) is now logged to aid debugging.

To be clear: this feature is not a bug and this PR is not the sole fix. Saving on room stop is intended behavior to prevent data loss, and admins must avoid forcibly killing the server while it is still cleaning up resources (at least for 30 seconds). This PR just minimizes the frequency of saving on room stop to reduce harm done to admins who have improperly configured deployments.

## Technical details

Adds `YRoomFileAPI.has_unsaved_changes` (backed by the existing save-scheduled flag) and gates the final save in `YRoom.stop()` on it. The stop drain applies updates after observers are removed, so it now marks the room dirty to keep those saved; a failed save re-arms the flag so it is retried rather than treated as clean.

## Testing

Added regression tests for the stop-save gate and the empty-notebook guard.

### DIFF
--- a/jupyter_server_documents/rooms/yroom.py
+++ b/jupyter_server_documents/rooms/yroom.py
@@ -1116,10 +1116,13 @@ class YRoom(LoggingConfigurable):
             # interruptible write that can corrupt the file if the server is
             # stopped uncleanly (e.g. a misconfigured environment).
             if not immediately and self.file_api.has_unsaved_changes:
+                self.log.info(f"Saving YRoom '{self.room_id}' on stop; it has unsaved changes.")
                 prev_jupyter_ydoc = self._jupyter_ydoc
                 self._save_task = asyncio.create_task(
                     self.file_api.save(prev_jupyter_ydoc)
                 )
+            elif not immediately:
+                self.log.info(f"Skipping redundant save-on-stop for YRoom '{self.room_id}'; no unsaved changes.")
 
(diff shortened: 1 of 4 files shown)


dlqqq:
GitHub actions is down... crap. https://www.githubstatus.com/

dlqqq:
CI failure is not related - needs investigation but it is happening on `main` as well.

dlqqq:
Proof of flaky test run: https://github.com/jupyter-ai-contrib/jupyter-server-documents/actions/runs/30501119236/job/90741415168

Issue #305, opened 2026-08-17
Notebook content lost after a room is freed and the client reconnects with divergent history (0.3.2)


albertmichaelj:
A student using my deployment of Juptyerhub with jupyter-server-documents (for RTC) reported data loss today, and I have been trying to chase down the cause with the help of Claude. I have the below issue that Claude wrote up following our back and forth, and I have read through it and agree that it describes the events. However, I fully acknowledge that this is AI generated output, and if you do not take AI generated issues, then I understand, and feel free to disregard this.

However, given that this was significant data loss (an entire notebook was lost). I wanted to raise the issue and see if there were any known mitigations or outstanding issues besides the issues that I found below (with the help of Claude). Thank you for your time!

## Summary

A student lost the content of a notebook after she closed her laptop and came back — she reopened
it and found a bare notebook with a single empty cell. The server was never restarted
(`restarts: 0`, up for 2h before and 1.5h after), and the most recent checkpoint of the file was
two days old, so the work between those points is gone.

We are on **0.3.2**, which already includes the divergent-sync fix from #254 / #258 (shipped in
0.3.0), so this is not that bug recurring — or if it is, the fix does not cover this path.

This looks closest to #257 (correctness of client-side divergent-history resolution) and possibly
#271. Filing separately with a full log because we have a clean, unambiguous reproduction trace and
I could not find one attached to either.

## Environment

| | |
|---|---|
| `jupyter_server_documents` | 0.3.2 |
| `jupyter_server` | 2.20.0 |
| `jupyterlab` | 4.6.2 |
| `jupyter-collaboration` | 5.0.0 (installed, `jupyter_server_ydoc` disabled) |
| `jupyter-ydoc` | 4.1.1 |
| `pycrdt` | 0.14.2 |
| Deployment | JupyterHub on Kubernetes (z2jh 4.4.1), one server per user |
| Home storage | NFS (AWS EFS) |
| Ingress | nginx, `proxy-read-timeout` at the 60s default at the time |

Single client throughout — no collaborators, no shared session.

## Timeline

Log is anonymised: username, hub name, IPs, file path and UUIDs replaced. Timestamps and ordering
are unmodified.

```
[I 10:51:35.113 ServerDocsApp] Freeing room 'json:notebook:<room-uuid>'.
[I 10:51:35.113 ServerDocsApp] Deleting YRoom 'json:notebook:<room-uuid>'.
[...]

Issue #12, opened 2025-05-05
Which version of JupyterLab to target


ellisonbg:
I wanted to raise the question about which version of JupyterLab to target. Some options:

* All releases in the 4.x series. This would be very nice, but not sure it is realistic as there may be other issues we run into.
* 4.4. The latest minor release has fixed the RTC prefix issue that we will most definitely run into.

My initial thinking is to target 4.4 initially and see how it goes. Thoughts?

Issue #11, opened 2025-05-05
Discussion: Debugging YDoc in browser side


jzhang20133:
Could use https://playcode.io/javascript as javascript playground. This helps us to understand a bit more about YDoc memory consumption and how different pattern of changing YDoc content impact its internal memory consumption, which pattern save memory and which pattern waste memory. 

Example debugging code to check YDoc, updates. 

```
import * as Y from 'yjs'
import * as util from "util";

const ydoc = new Y.Doc();

// Create a top level data type
const ytext = ydoc.getText("listOfNumbers");

// Insert from right to left to avoid optimizations
ytext.insert(0, "abcfiehwighewgweg");
ytext.delete(0, "abcfiehwighewgweg");
ytext.insert(0, "894uug9j4983gh4893ug89");

console.log(ytext.toString())
console.log(ytext)
console.log(ydoc)
// Log the underlying edits
const updates = Y.encodeStateAsUpdate(ydoc);
console.log(Y.decodeUpdate(updates))
console.log(
  util.inspect(Y.decodeUpdate(updates), { colors: true, depth: null })
);
```

PR #310, opened 2026-09-01
Survive messages from stale clients in the room message queue

### LINKED ISSUE #271: A disconnected client may permanently poison its room's message queue
A client that has already left a room can leave a stale `SYNC_UPDATE` frame in that room's message queue. When the room's queue task pops that frame, it crashes and from that moment the room silently stops processing messages for all other clients. New clients can still connect, but their `SYNC_UPDATE` frames are never applied to the doc or broadcast. There is no error surfaced to any client. Related #267 (likely different root cause)

There are two layers here.

### 1. `Clients.get()` raises `UnboundLocalError` for an absent client

`jupyter_server_documents/websockets/clients.py:133-145` — when `client_id` is in neither `self.desynced` nor `self.synced`, the local `client` variable is never assigned, so the check on line 141 raises `UnboundLocalError` before the intended `Exception(error_message)` on line 145:

### 2. The call chain from the queue task to `get()` has no exception boundary, so any exception raise kills the room

```
_process_message_queue   -> handle_maessage -> handle_sync_update -> _should_ignore_update -> Clients.get() [has no exception handling]
```

Every subsequent message queued for that room is then stuck if there is an exception processing that message

### Steps to reproduce

Sequence:
1. Client A: handshake, burst 20 `SYNC_UPDATE` frames, close abruptly.
2. Server: `_process_message_queue` pops a queued update, calls `Clients.get(client_id)` — client A already removed from both groups and encounters `UnboundLocalError`
3. Client B connects to the same room, sends `SyncStep1`, never receives the server's `SyncStep2` reply,  handshake times out. 

Traceback matches:
```
Task exception was never retrieved
future: <Task ... _process_message_queue() ... exception=UnboundLocalError(...)>
  File ".../rooms/yroom.py", line 788, in handle_sync_update
    if self._should_ignore_update(client_id, "SyncUpdate"):
  File ".../rooms/yroom.py", line 931, in _should_ignore_update
    client = self.clients.get(client_id)
  File ".../websockets/clients.py", line 141, in get
    if client.websocket and client.websocket.ws_connection:
UnboundLocalError: cannot access local variable 'client' where it is not associated with a value
```

### Proposed fix

```
def get(self, client_id: str) -> YjsClient:
    client = self.desynced.get(client_id) or self.synced.get(client_id)
    if client is None or not (client.websocket and client.websocket.ws_connection):
        error_message = f"The client_id '{client_id}' is not found in client group in room '{self.room_id}'"
        self.log.error(error_message)
        raise Exception(error_message)
    return client
```

and 

```
client_id, message = queue_item
try:
    await self.handle_message(client_id, message)
except Exception:
    self.log.exception(
        f"Error handling message from client '{client_id}' in room "
        f"'{self.room_id}'; skipping it and continuing."
    )
finally:
    self._message_queue.task_done()
```
The latter may cover #267

stebo85:
Fixes #271.

## The bug

A message enqueued by a client that disconnects before the message is processed crashes the room in two compounding ways:

1. **`YjsClientGroup.get()` raises `UnboundLocalError`.** The lookup assigns `client` only inside `if client_id in self.desynced` / `if client_id in self.synced` blocks, so when the client ID is in neither dict (it was already removed), the following `if client.websocket ...` references an unassigned local and raises `UnboundLocalError` instead of the intended "client not found" error.

2. **The exception escapes `YRoom._process_message_queue()`.** The queue loop awaits `handle_message()` with no exception guard, so the raised error terminates the background task. From that point on, the room is permanently poisoned: it stays connected, but no message from *any* client is ever processed again (reported in #271 against chat rooms). Additionally, because `task_done()` is skipped for the failed message, `Queue.join()` can never unblock.

## Reproduction

Both halves reproduce without a browser, straight against the classes:

```python
import asyncio, logging
from jupyter_server_documents.websockets import YjsClientGroup
from jupyter_server_documents.rooms.yroom import YRoom

# (1) lookup of a client that has already been removed
group = YjsClientGroup(room_id="text:file:demo", log=logging.getLogger("demo"))
group.get("no-such-client")
# -> UnboundLocalError: cannot access local variable 'client' ...
#    (expected: the "client_id ... is not found" exception)

# (2) one raising message permanently stops the room's queue
class Room:
    _process_message_queue = YRoom._process_message_queue
    room_id, log, handled = "demo", logging.getLogger("demo"), []
    def __init__(self): self._message_queue = asyncio.Queue()
    async def handle_message(self, client_id, message):
        if client_id == "gone": raise RuntimeError("client already removed")
        self.handled.append(client_id)

async def main():
    room = Room()
    room._message_queue.put_nowait(("gone", b""))   # e.g. a frame from a client that disconnected
    room._message_queue.put_nowait(("alive", b""))
    task = asyncio.create_task(room._process_message_queue())
    await asyncio.wait_for(room._message_queue.join(), timeout=2)  # hangs: task_done() never ran
    print(room.handled)

asyncio.run(main())
# -> TimeoutError; task is dead after the first message, "alive" is never handled.
#    After the fix: join() returns and handled == ["alive"].
```

In a real deployment the trigger is a client that sends an update and disconnects before the queue drains (a closed tab, a laptop lid): `handle_sync_update()` → `_should_ignore_update()` → `clients.get()` raises, and every later message from *every* client in that room is silently ignored (the chat-room symptom in #271).

## The fix

- `YjsClientGroup.get()` now uses a `.get()`-based lookup (`self.synced.get(client_id) or self.desynced.get(client_id)`, preserving the existing synced-over-desynced preference) and raises the intended, descriptive exception when the client is unknown or its socket is gone.
- `YRoom._process_message_queue()` wraps `handle_message()` in `try/except`: a failure is logged with `log.exception()` and the message is skipped, and `task_done()` moved into a `finally` block so `Queue.join()` cannot deadlock.

## How it was tested

New tests, both failing before the fix and passing after:

- `jupyter_server_documents/tests/test_client_group.py`: `get()` returns desynced/synced clients; `get()` on an unknown or removed client ID raises the intended exception rather than `UnboundLocalError` (regression tests for #271).
- `jupyter_server_documents/tests/test_yroom.py::TestYRoomMessageQueue`: drives `YRoom._process_message_queue()` with a handler that raises for the first message, and asserts that the next message is still processed, that `Queue.join()` unblocks (i.e. `task_done()` ran for the failed message too), and that the background task keeps running until halted via the `None` sentinel.

Full Python suite: `python -m pytest jupyter_server_documents/tests` — 176 passed, 4 skipped (the skips are pre-existing and unrelated).

## Relation to #305

The analysis in #305 notes that the single-slot pending-SyncStep2 future (`_pending_ss2_future` / `_pending_ss2_client_id`) is related territory: replies from two tabs on one document can orphan each other. This PR deliberately does **not** touch the handshake path — it only fixes the crash-on-stale-client lookup and the unguarded queue loop — so it composes with the per-client futures / late-SyncStep2 changes proposed there and should apply cleanly before or after them.

Found while integrating downstream.

Analysis and proposed fix were discussed in #271 before opening this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


shravan-achar:
The changes may have broken the tests in https://github.com/jupyter-ai-contrib/jupyter-server-documents/tree/main/ui-tests/tests. Please take a look

dlqqq:
FYI @shravan-achar I have given you release rights to the repo :) 

Would love for you to help maintain this package since you all use it -- I think you would be a great owner for it.

Feel free to merge PRs and publish releases -- @Zsailer can help walk you through the process. You already have the rights 👍

Zsailer:
Hey all! I'll work with @shravan-achar to review this PR (and the other ones, #11, #12, and #13) over the next week. 

It's a lot of code to review, so it will take us a few weeks. Really appreciate your patience!

stebo85:
> The changes may have broken the tests in https://github.com/jupyter-ai-contrib/jupyter-server-documents/tree/main/ui-tests/tests. Please take a look

CI finally ran on #310. Everything passes except Integration tests: 6 failed / 27 passed, all in chat-router.spec.ts (2 unique tests × 3 retries).

It's pre-existing. [#316](https://github.com/jupyter-ai-contrib/jupyter-server-documents/pull/316) — a dependabot PR that changes only ui-tests/yarn.lock — fails with the identical signature: same two tests, same 6/27. 

The router never fires because an observer callback raises first:

```
yroom.py:1184 _finalize_stop → drain_observer_removals → pycrdt transaction __exit__
  jupyterlab_chat/ychat.py:367 _dispatch_message_event
    message = Message(**item.to_py())
AttributeError: 'NoneType' object has no attribute 'to_py'
```

item is None inside jupyterlab_chat's message dispatcher. Both candidates moved inside the window and neither is upper-bounded: jupyterlab_chat 0.23.1 → 0.25.0 (0.25.0 published 3 Sep, the day these runs executed) and pycrdt 0.14.2 → 0.14.4 (pycrdt>=0.12.49).

Note our run was queued 1 Sep but only executed 3 Sep once approved — which is why it picked up a package released that morning.

jupyter_server_documents/tests/test_yroom.py
```diff
@@ -36,6 +36,64 @@ async def test_on_stop_callbacks(self, make_yroom: MakeYRoom):
...
+        terminates the `_process_message_queue()` background task nor skips
+        the `task_done()` call for the failed message.
+
+        Regression test for #271: a queued message from a client that had
+        already disconnected raised from `handle_message()`, which terminated
+        the background task and left the room unable to process any future
+        message from any client.
+        """
+        import logging
+        from jupyter_server_documents.rooms.yroom import YRoom
```
Zsailer: Can we move these local imports to the top of the file? This is a very AI-generated signature 🙂

Zsailer:
@stebo85 this PR looks solid. I left a minor comment on the unit test. Otherwise, this looks good to me. I'm happy to take over the PR, make the minor change, merge and release. Otherwise, I can wait for any changes you'd like to make yourself.

stebo85:
Dear @Zsailer - thank you so much! The import is now fixed. Will also fix in my other pull requests. Rebasing now and then ready to go from my side.

PR #321, opened 2026-09-10
[Feature] Bind documents to their kernel lazily at execute time

Zsailer:
Server-side execution needs a document's room to hold a kernel client. Instead of establishing that document↔kernel mapping through a custom session manager, this binds it lazily: the execute route already carries both the kernel id (in the URL) and the document id (in the body), so `POST /api/kernels/{kernel_id}/execute` connects the room to that kernel on demand — and rebinds if the document later points at a different kernel.

- **Disconnect** is tied to the room's lifecycle (a stop-callback), not to a session.
- **Restart / interrupt / shutdown** are handled by a `kernel_actions/v1` event listener in the extension app. (A restart keeps the same kernel id, so the lazy binding can't detect on its own that the client went stale.)

With this, `YDocSessionManager` is removed and the stock jupyter_server `SessionManager` works unchanged.

The room builds its client from the kernel manager's `client_factory` trait (overridable to select a client class per transport) rather than `kernel_manager.client()`, which clones the manager's `Session` and gives every client the same ZMQ DEALER identity, misrouting execute replies. Channels start with `hb=False`; readiness is confirmed by a `kernel_info` round-trip, which also avoids a heartbeat-thread `ZMQError` when a kernel's context is torn down on shutdown.


Zsailer:
Breaking this PR into three smaller, more digestible PRs.

Zsailer:
Split into three smaller, layered PRs for easier review (together they reconstruct exactly this change):

1. #323 — order execute requests by a monotonic per-client sequence
2. #324 — run cells via a manual execute + iopub loop (stacked on #323)
3. #326 — remove YDocSessionManager; bind documents to kernels lazily (stacked on #324)

PR #323, opened 2026-09-11
[Feature] Order execute requests by a monotonic per-client sequence

Zsailer:
Cells execute server-side via `POST /api/kernels/{kernel_id}/execute`. A "Run All" (or several quick runs) fires multiple of these POSTs, and they can reach the server out of order — HTTP/2 streams and independent HTTP/1.1 connections make no ordering promise — so the server needs to re-establish the user's intended order. Today it does that with a `previous_request_id` predecessor chain; this replaces that with a per-client monotonic sequence, which gives a simpler client contract and, because it's an absolute position rather than a relative link, lets the server detect gaps/duplicates/rewinds and recover deterministically.

### REST request model

```jsonc
POST /api/kernels/{kernel_id}/execute
{
  "document_id": "string",     // required — room / document id
  "cells": [                   // required — run atomically, in order
    { "cell_id": "string", "source_hash": "string" }
  ],

  // ordering (client_id is required when sequence is present)
  "client_id": "string",       // identifies the browser tab; scopes `sequence`
  "sequence": 0,               // monotonic counter per client_id, starts at 0
  "request_id": "string"       // opaque UUID for logging / tracing
}
```

What changed in the ordering fields:

```diff
   "client_id": "string",
-  "request_id": "string",
-  "previous_request_id": "string"  // wait for this request to be enqueued first
+  "sequence": 0,                   // monotonic counter per client_id, starts at 0
+  "request_id": "string"           // opaque UUID for logging / tracing
```

The server enqueues per `client_id` in `sequence` order, buffering out-of-order arrivals until their predecessors show up. Because `sequence` is an absolute position, the server can reject a rewind (`sequence` below what it has already seen) and treat `sequence == 0`, when it is already ahead, as a session reset (new tab, kernel swap, explicit reset): state is cleared and pending waiters are abandoned with `SessionResetError` (surfaced as `409`, which the frontend retries from `sequence=0`). A predecessor that never arrives is still capped by a timeout — same safety net as before — but recovery is the deterministic reset rather than proceeding out of order. An in-place kernel restart keeps the same kernel id, so the client's counter stays valid — `disconnect_kernel(reset_sequences=False)` preserves it across that reconnect.

First of three layered PRs; this one is just the ordering change (the execute-loop rewrite and the session-manager simplification follow on top).

### DIFF
--- a/jupyter_server_documents/execution_handlers.py
+++ b/jupyter_server_documents/execution_handlers.py
@@ -3,7 +3,12 @@ from jupyter_server.base.handlers import APIHandler
 from tornado import web
 from tornado.escape import json_encode
 
-from .rooms.ynotebook_room import YNotebookRoom, SourceMismatchError, PredecessorTimeoutError
+from .rooms.ynotebook_room import (
+    YNotebookRoom,
+    SourceMismatchError,
+    SequenceOutOfRangeError,
+    SessionResetError,
+)
 
 
 AUTH_RESOURCE = "executions"
@@ -32,9 +37,9 @@ class KernelExecuteHandler(ExecutionsAPIHandler):
       ],
 
       // Execution ordering (optional)
-      "client_id":          "string",  // document client ID
-      "request_id":         "string",  // UUID for this request
-      "previous_request_id":"string"   // wait for this request to be enqueued first
+      "client_id": "string",  // identifies the browser tab; scopes `sequence`
+      "sequence":  0,         // monotonic counter per client_id, starts at 0
+      "request_id":"string"   // opaque UUID for logging / tracing
     }
     ```
 
@@ -51,8 +56,8 @@ class KernelExecuteHandler(ExecutionsAPIHandler):
     ## Responses
     - ``200 null``  — accepted (fire-and-forget)
     - ``400``       — bad request
-    - ``408``       — predecessor request timed out
     - ``409 {"error": "source_mismatch", "cell_id": "..."}`` — source diverged
+    - ``409``       — session reset (client should clear its counter and retry with sequence=0)
     """
 
     @web.authenticated
@@ -70,7 +75,13 @@ class KernelExecuteHandler(ExecutionsAPIHandler):
 
         client_id = body.get("client_id")
         request_id = body.get("request_id")
-        previous_request_id = body.get("previous_request_id")
+        sequence = body.get("sequence")
+
+        if sequence is not None:
+            if not isinstance(sequence, int) or sequence < 0:
+                raise web.HTTPError(400, "sequence must be a non-negative integer")
+            if not client_id:
+                raise web.HTTPError(400, "client_id is required when sequence is present")
 
         yroom = self.settings["yroom_manager"].get_room(document_id)
         if yroom is None:
@@ -83,14 +94,19 @@ class KernelExecuteHandler(ExecutionsAPIHandler):
                 cells_payload,
                 clear_outputs=True,
                 request_id=request_id,
-                previous_request_id=previous_request_id,
+                client_id=client_id,
+                sequence=sequence,
             )
         except SourceMismatchError as e:
             self.set_status(409)
             self.finish(json_encode({"error": "source_mismatch", "cell_id": e.cell_id}))
             return
-        except PredecessorTimeoutError:
-            raise web.HTTPError(408, "Timed out waiting for previous_request_id to be enqueued")
+        except SessionResetError:
+            self.set_status(409)
+            self.finish(json_encode({"error": "session_reset"}))
+            return
+        except SequenceOutOfRangeError as e:
+            raise web.HTTPError(400, str(e))
         except (LookupError, ValueError, RuntimeError) as e:
             raise web.HTTPError(400, str(e))
 
--- a/jupyter_server_documents/rooms/ynotebook_room.py
+++ b/jupyter_server_documents/rooms/ynotebook_room.py
@@ -21,10 +21,6 @@ if TYPE_CHECKING:
     from ..outputs.output_processor import OutputProcessor
 
 
-# How long a request with previous_request_id will wait for its predecessor.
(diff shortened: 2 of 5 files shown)


Zsailer:
Heads up: the Integration tests failure looks unrelated to this PR. The only failures are two tests in `tests/chat-router.spec.ts` ("each chat message fires the router exactly once" and "reconnection alone does not re-fire the router") — chat-message routing/reconnection, whereas this PR only touches execute-request ordering. The notebook/execution UI tests passed (27 passed), and this branch is based directly on `main`, so it looks like pre-existing flakiness rather than anything here. Happy to re-run the job if that's useful.

Zsailer:
@krassowski would love your high level look at minor change to the REST API model here.

albertmichaelj:
Thanks for this, @Zsailer. We run server-side execution in production on a JupyterHub for a university course, where laptops sleep mid-session and rooms get freed and recreated under open tabs (see #305). We patched a few `previous_request_id` chain failures client-side in a fork. This PR fixes the design underneath them, so we'd rather support it than open a competing PR. Below: what it fixes for us, one remaining gap, two small things, and tests we can contribute.

### 1. What this fixes for us

- **A kernel swap stalls the next run for 10s, then 408s.** We reproduced this on demand with v0.3.3's chain: shut the kernel down, then run a cell. The server clears its enqueued-request history when the room gets a new kernel, but the client still names a predecessor from that history. Under Run All every cell chains behind that one, so the whole batch is lost. Here, ``const docKey = `${documentId ?? path}:${clientId}:${kernelId}`;`` starts a new kernel at sequence 0, and `disconnect_kernel(reset_sequences=False)` keeps the counter valid across an in-place restart. That is the same distinction we ended up drawing.
- **A chain that stayed broken until a page reload.** Every execute returned 408 for about 11 minutes. Restarting or replacing the kernel didn't help, because the bad state was in the browser. That was our fork's client bug, not main's, but it showed how much the old protocol relied on the client recovering perfectly. Here the server bounds every wait itself:
  - `if sequence == 0 and current > 0:` resets the counter and wakes all waiters.
  - A waiter that times out runs `self._seq_generation[client_id] += 1` and `self._seq_cv.notify_all()`, so its peers fail with it instead of each waiting out a timeout in turn.
  - Every failed response except `source_mismatch`, where the server has already consumed the slot, runs `nextSeqByDoc.delete(docKey)`.

  A client can no longer get stuck: the worst case is one timed-out wait, then a working click.

### 2. Remaining gap: a room freed while the tab sleeps

A room freed and recreated during sleep keeps the same `docKey`, but the new room starts with `_next_seq = {}`. With #326's lazy binding, the first request after wake reaches the sequence check with `sequence = N > 0` against `current = 0`. That isn't a reset, so it waits in `while self._next_seq.get(client_id, 0) < sequence:` until `kernel_info_timeout` (60s by default), then gets a 409 `session_reset`. That's `test_lost_predecessor_times_out` at production timeout, six times the 10s it replaces, and a Run All in this state loses every cell. A unit check against 1708091 confirms it: sequences 7–11 sent into a fresh room all raised `SessionResetError` together at the deadline, and `sequence=0` then enqueued immediately.

The server can't tell this apart from ordinary reordering (sequence 1 reaching a new room before sequence 0), so the fix belongs in the client: drop the counter, or fold a connection epoch into `docKey`, after a reconnect that follows a long absence. Two lessons from our fork:

- **Measure absence from a "last alive" stamp refreshed on an interval, not from the `'disconnected'` event.** No JS runs while a laptop sleeps, so that event is only delivered at wake, seconds before the reconnect. Measured from it, a 20-minute lid close looks like a 3s blip.
- **Gate the reset.** A `sequence == 0` reset abandons every waiter for that client, so resetting after every network blip would drop the in-flight requests the sequence exists to order. We reset after 45s of absence (the default `inactivity_timeout` is 60s) and refresh the stamp every 10s. It's best-effort: room inactivity counts from the last edit or awareness change, not from when the tab left, so a shorter absence can still straddle a cleanup sweep. When it misses, the cost is one timed-out wait, as now.

We can open this as a small follow-up on top of the stack if you'd like.

### 3. Two small things

- **The `catch` block doesn't reset the counter.** If a POST fails at the network layer and never arrived, the next run waits the full timeout. Calling `nextSeqByDoc.delete(docKey)` in the `catch` is cheap either way: `sequence 0` never waits, and the only waiters it can abandon are this tab's own.
- **The reset message always blames the kernel.** `'Cell not executed: the kernel changed while the request was in flight. Please re-run the cell.'` also appears after a timeout (`predecessor never arrived`), when the kernel is fine. Neutral wording would stop users restarting a healthy kernel. Nits: the `execute_cells` docstring still says "there is no timeout heuristic", and the PR description says the frontend retries a 409 from `sequence=0`, which reads as automatic; the code clears the counter and asks the user to re-run.

### 4. Tests we can contribute

Ours are in albertmichaelj/jupyter-server-documents@b900bbc, albertmichaelj/jupyter-server-documents@808ec19 (the liveness-stamp part), albertmichaelj/jupyter-server-documents@6799af7 and albertmichaelj/jupyter-server-documents@7788773. They target our old chain, but they port once the counter bookkeeping moves into a small helper. #323 has no client-side tests yet.

- `yprovider.spec.ts`: `bumps the epoch only when downtime could have outlived the room` (a 2s blip keeps the epoch, 45s bumps it, and repeated `'disconnected'` events during retries don't refresh the stamp) and `bumps after a sleep even though the disconnect is only observed at wake` (a 20-minute gap first seen at wake still bumps).
- `executionChain.spec.ts`: three `shouldBumpEpoch` cases, `recovers after a Run All burst in which every request fails`, `starts a fresh chain when the notebook gets a new kernel`, and `survives a restart, which keeps the same kernel id`.
- Server side: the burst check from section 2, as a variant of `test_lost_predecessor_times_out`.

Tell us which form you'd like and we'll adapt them.

Drafted with the help of Claude Code (Claude Opus 5.5) and reviewed by me before posting.

PR #324, opened 2026-09-11
[Feature] Run cells via a manual execute + iopub loop

Zsailer:
`_run_item` currently runs a cell with `client._async_execute_interactive(..., output_hook=...)`. That helper drives its own receive loop with `zmq.Poller.poll`, which doesn't cooperate with the asyncio event loop. This swaps it for an explicit `client.execute()` plus draining the async iopub channel (`get_msg`, which uses `zmq.asyncio` and yields properly), filtering on `parent_header.msg_id` until we see `status: idle` for our request.

No behavioural change to what lands in the YDoc (execution_state / execution_count / outputs are written exactly as before); this is purely how the messages are driven.

---

Stacked on #323 (monotonic sequence ordering) — its commit shows here until that merges; the change to review here is the `_run_item` rewrite. The session-manager simplification follows on top.

### DIFF
--- a/jupyter_server_documents/execution_handlers.py
+++ b/jupyter_server_documents/execution_handlers.py
@@ -3,7 +3,12 @@ from jupyter_server.base.handlers import APIHandler
 from tornado import web
 from tornado.escape import json_encode
 
-from .rooms.ynotebook_room import YNotebookRoom, SourceMismatchError, PredecessorTimeoutError
+from .rooms.ynotebook_room import (
+    YNotebookRoom,
+    SourceMismatchError,
+    SequenceOutOfRangeError,
+    SessionResetError,
+)
 
 
 AUTH_RESOURCE = "executions"
@@ -32,9 +37,9 @@ class KernelExecuteHandler(ExecutionsAPIHandler):
       ],
 
       // Execution ordering (optional)
-      "client_id":          "string",  // document client ID
-      "request_id":         "string",  // UUID for this request
-      "previous_request_id":"string"   // wait for this request to be enqueued first
+      "client_id": "string",  // identifies the browser tab; scopes `sequence`
+      "sequence":  0,         // monotonic counter per client_id, starts at 0
+      "request_id":"string"   // opaque UUID for logging / tracing
     }
     ```
 
@@ -51,8 +56,8 @@ class KernelExecuteHandler(ExecutionsAPIHandler):
     ## Responses
     - ``200 null``  — accepted (fire-and-forget)
     - ``400``       — bad request
-    - ``408``       — predecessor request timed out
     - ``409 {"error": "source_mismatch", "cell_id": "..."}`` — source diverged
+    - ``409``       — session reset (client should clear its counter and retry with sequence=0)
     """
 
     @web.authenticated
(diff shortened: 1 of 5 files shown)


albertmichaelj:
Thanks for this, @Zsailer. We run jupyter-server-documents in production for a university course, and our CI hit the failure this PR removes. Our fork fixed it by carrying a patched copy of jupyter_client's execute loop. Owning the loop, as you do here, is the smaller fix and looks like the right long-term shape. Two notes from that work, both checked against 79c3791 (still the head today) with jupyter_client 8.10.0 and ipykernel 7.3.0.

**1. This also fixes lost outputs.** `_async_execute_interactive` waits on the raw iopub socket, then reads from the channel with a zero timeout:

```python
events = dict(await poller.poll(timeout_ms))
...
msg = await ensure_async(self.iopub_channel.get_msg(timeout=0))
```

The channel's `get_msg` runs its own poll and raises `queue.Empty` if that poll finds nothing:

```python
ready = await self.socket.poll(timeout_ms)
if ready:
    res = await self._recv()
    return res
else:
    raise Empty
```

If the second poll misses what the first one reported, `Empty` escapes after the `execute_request` has already gone out. The kernel runs the cell, the rest of its outputs are never read, and `_run_item` marks the cell idle. We hit this in this repo's `console-for-notebook` UI test, on the first run against a freshly started kernel. The log shows `execute_cell error cell_id=…: ` with nothing after the colon, because `str(queue.Empty())` is empty.

`msg = await iopub.get_msg(timeout=None)` has no second zero-timeout read: it waits on the socket and reads from it, so that window is gone. One correction for the description: in jupyter_client 8.9.1 and 8.10.0 the helper already uses `zmq.asyncio.Poller()` and awaits it, so it does yield. The read-after-poll race is the concrete reason to own the loop. Saying so in the PR body would stop anyone "simplifying" back to the helper later.

**2. The `execute_reply` is never read.** The helper ends with `return await self._async_recv_reply(msg_id, timeout=timeout)`. The new loop breaks on `status: idle`, and nothing else reads the room client's shell channel after `_fetch_kernel_info()`. So each cell's reply stays queued on that client's shell socket, and the room never sees the kernel's verdict (`ok`/`error`/`aborted`). iopub can't show an abort: ipykernel's `dispatch_shell` calls `self._send_abort_reply(stream, msg, idents)` then `self._publish_status_and_flush("idle", "shell", stream)`, so iopub gets only busy and idle. The replies matter to the room already: `_connect_client`'s docstring avoids `kernel_manager.client()` because "The kernel's ROUTER then routes execute_reply to the wrong socket." A one-line fix after the loop:

```python
            reply = await client._async_recv_reply(msg_id, timeout=None)
```

It already filters on `parent_header.msg_id`. ipykernel sends the reply before it publishes idle, so this normally adds no wait. The three `execute = MagicMock(return_value="req-1")` stubs in `test_yroom_kernel.py` then each need `room._kernel_client._async_recv_reply = AsyncMock(return_value={"content": {"status": "ok"}})`. Without it, `test_execute_input_writes_execution_count` fails. With it, the suite plus the two tests below gave 175 passed, 4 skipped.

**Tests we can offer**, in your mock style (a `MagicMock` client, with the worker running `_run_item`):

- `test_outputs_survive_a_zero_timeout_read_miss`: the iopub channel wraps a real zmq PULL socket, jupyter_client's own `_async_execute_interactive` is bound onto the client, and the first zero-timeout read raises `Empty`. It **fails on `main`** (the `execute_result` is lost) and passes here.
- `test_execute_reply_is_consumed`: asserts `_async_recv_reply` is awaited with the request's msg_id. It passes on `main`, **fails on this PR as-is**, and passes with the line above.

Both are adapted from `TestResilientExecuteInteractive` in `jupyter_server_documents/tests/test_room_lifecycle_guards.py` at albertmichaelj/jupyter-server-documents@7788773. Happy to push them to a branch on top of this one, or paste them here. If the reply status later drives stop-on-error, the same commit has a `TestStopOnError` class in `test_yroom_kernel.py` built on it.

Aside: #323's guard (`await sessionContext.ready;` then `if (!kernelId) { … return false; }`) matches one we added independently after our CI logs showed execute POSTs for kernel `undefined`. Nothing to add there.

Drafted with the help of Claude Code (Claude Opus 5.5) and reviewed by me before posting.

